PR #42, opened 2018-03-10
Fix reCaptcha being passed by bots

### LINKED ISSUE #3: Rework captcha support
bug

**What happened**:
Even with the captcha enabled, spam requests still come in.

All spam form submissions shouldn't be possible.

Enable the captcha and wait for spam.

**Environment**:
- Syna Theme version:  v0.7.0
- Hugo version:  0.30.2

mpourismaiel:
reCaptcha was not implemented like the [reCaptcha guidelines](https://developers.google.com/recaptcha/docs/invisible) and I think that caused the issue #3 

Also the message regarding lack of reCaptcha on the page was not being removed if reCaptcha did not appear immediately after render due to internet speed. This PR fixes that as well.

**Release note**:

```release-note

```


stp-ip:
This change seems to break:
* css to hide the recaptcha notice
* disabled button so clicking is prevented until data is checked via js
* data checking

The js change seems good so.

mpourismaiel:
I'm working on the problems but I don't understand "css to hide the recaptcha notice"

stp-ip:
Recaptcha can be inlined and the inlined recaptcha icon can be hidden. We did hide it before, but the other option is to not hide it, but design it better into the website.

mpourismaiel:
If everything else is okay I can get to styling it and add an option for hiding it perhaps.

stp-ip:
Recaptcha and form looks good now.

Still needs styling and conflict resolution. The errors not showing is already open as issue.

mpourismaiel:
Shall I fix the error messages in the same PR? Also how should I style the recaptcha?

PR #46, opened 2018-03-12
Remove attribution from footer

### LINKED ISSUE #39: Check if attribution is set to false and then also remove it
bug

**What happened**:
Currently attribution in the footer is not removed, when set to "false" in config.toml. It needs to be removed.

Setting it to "false" or removing should work.

lauragift21:
(#39)

Attribution still displays even with false assigned to it. Fix for this is to completely remove it from config.toml

**Release note**:

```release-note

```

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -12,6 +12,7 @@ googleAnalytics = ""
 [params]
   name = "Okkur Labs"
   description = "Open Source Theme for your next project"
+  attribution = "false"
 
   # Branding options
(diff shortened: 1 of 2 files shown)


stp-ip:
That also works, but not completely the idea.
We want the attribution to be available, but want another check added.

The result should be, that the user can either remove the attribution config or set it to "false". Both should not display the attribution.

If it's set to "true" it should show the attribution.

stp-ip:
Looks good. Great.

PR #49, opened 2018-03-14
Add baseof template

### LINKED ISSUE #29: Make use of baseOf.html
feature

**What happened**:
Currently navigation, footer are partials. These should be moved to baseOf.html.

Use baseOf.html.

https://gohugo.io/templates/base/#define-the-base-template

mpourismaiel:
Adds `baseof.html` layout so some more used partials can settle in there.

fixes #29

I'm not sure if this is the implementation #29 needs. Let me know if this is not the change you're looking for.

**Release note**:

```release-note

```

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -1,0 +1,16 @@
+<html lang="{{ .Site.LanguageCode | default (printf "en-us") }}">
+
+  {{ partial "head.html" . }}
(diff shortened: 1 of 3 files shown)


stp-ip:
Awesome. That simplifies the layout files a bit and let us use the best practice from Hugo.

lauragift21:
Awesome work. :+1:

PR #51, opened 2018-03-14
Add support for netlify form-handling

### LINKED ISSUE #43: Netlify form submission support
feature

**What happened**:
Currently we support self hosted email sends and formspree.

Support netlify contact form submissions.

https://www.netlify.com/docs/form-handling/

mpourismaiel:
Adds support for netlify form-handling

fixes #43 

This might not be a good solution since it doesn't take any global configuration into account

**Release note**:

```release-note

```

### DIFF
--- a/layouts/partials/fragments/contact.html
+++ b/layouts/partials/fragments/contact.html
@@ -35,8 +35,7 @@
       <div class="col-12">
         <form class="contact" action="
           {{- if not $contact.netlify -}}
-            {{- $postEmailUrl := printf "https://formspree.io/$%s" $contact.email }}
-            {{- $postUrl := $contact.posturl | default $postEmailUrl -}}
+            {{- $postUrl := $contact.posturl | default (printf "https://formspree.io/%s" $contact.email) -}}
             {{- printf $postUrl -}}
           {{- end -}}
         " method="POST" name="
--- a/layouts/partials/fragments/contact.html
+++ b/layouts/partials/fragments/contact.html
@@ -33,11 +33,18 @@
     </div>
     <div class="row">
       <div class="col-12">
-        {{ if not $contact.posturl }}
-          <form class="contact" action="https://formspree.io/{{ $contact.email }}" method="POST" name="sentMessage" id="{{ $contact.formName | default "contactForm" }}">
-        {{ else }}
-          <form class="contact" action="{{ $contact.postURL }}" method="POST" name="sentMessage" id="{{ $contact.formName | default "contactForm" }}">
-        {{ end }}
+        <form class="contact" action="
+          {{- if not $contact.netlify -}}
+            {{- $postUrl := $contact.posturl | default $contact.email -}}
+            {{- printf "https://formspree.io/%s" $postUrl -}}
+          {{- end -}}
+        " method="POST" name="
+        {{- if $contact.netlify -}}
+          netlify
+        {{- else -}}
+          sentMessage
+        {{- end -}}
+        " id="{{ $contact.formName | default "contactForm" }}">
             <div class="row py-3">
               <div class="col-md-6">
                 {{ with $contact.fields.name }}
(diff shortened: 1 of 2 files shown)


layouts/partials/fragments/contact.html
```diff
@@ -33,11 +33,18 @@ <h2>{{- . | markdownify -}}</h2>
...
       <div class="col-12">
-        {{ if not $contact.posturl }}
-          <form class="contact" action="https://formspree.io/{{ $contact.email }}" method="POST" name="sentMessage" id="{{ $contact.formName | default "contactForm" }}">
-        {{ else }}
-          <form class="contact" action="{{ $contact.postURL }}" method="POST" name="sentMessage" id="{{ $contact.formName | default "contactForm" }}">
-        {{ end }}
+        <form class="contact" action="
+          {{- if not $contact.netlify -}}
+            {{- $postUrl := $contact.posturl | default $contact.email -}}
+            {{- printf "https://formspree.io/%s" $postUrl -}}
```
stp-ip: This would always use formspree afaik. Even with `postUrl = "mailout.example.com"` it would use `https://formspree.io/mailout.example.com`

stp-ip:
Small issue I think, but in general looks good.

layouts/partials/fragments/contact.html
```diff
@@ -33,11 +33,18 @@ <h2>{{- . | markdownify -}}</h2>
...
       <div class="col-12">
-        {{ if not $contact.posturl }}
-          <form class="contact" action="https://formspree.io/{{ $contact.email }}" method="POST" name="sentMessage" id="{{ $contact.formName | default "contactForm" }}">
-        {{ else }}
-          <form class="contact" action="{{ $contact.postURL }}" method="POST" name="sentMessage" id="{{ $contact.formName | default "contactForm" }}">
-        {{ end }}
+        <form class="contact" action="
+          {{- if not $contact.netlify -}}
+            {{- $postUrl := $contact.posturl | default (printf "https://formspree.io/%s" $contact.email) -}}
+            {{- printf $postUrl -}}
```
stp-ip: What's the reason for the additional printf instead of just plain:
`{{ $contact.posturl | default (printf "https://formspree.io/%s" $contact.email) }}`
I'm fine with both, just wanted to get the reason behind that.

stp-ip:
Another thing to consider. Should we use netlify's recaptcha support, when using netlify? Might be easier for people that just want to use netlify.

stp-ip:
Netlify does not parse the form perfectly so a missing whitespace before `netlify>` in the generated form will fail.

Also formName needs to be formname <-- I come to really hate that hugo case

My suggestions would be something like:
```
        <form class="contact" method="POST" name="{{ $contact.formname | default "contactForm" }}" id="contactForm
          {{- if $contact.netlify -}}
            " netlify
          {{- else -}}
            " action="{{- $contact.posturl | default (printf "https://formspree.io/%s" $contact.email) -}}"
          {{- end -}}
          >
```

stp-ip:
Another way would be to not remove the whitespace via the `-`, but that feels hacky.

layouts/partials/fragments/contact.html
mpourismaiel: It was left behind from the previous bug I wrote. Sorry.

mpourismaiel:
I added your code and it works very well. The only issue is syntax highlighting of the code. Check this out:

screenshot-2018-03-15_21-58-35

Regarding the other issue, I was hoping to add that when #3 is fixed since adding Netlify captcha support would mean that we should disable our own captcha and perhaps remove our error for when captcha can not be loaded.

stp-ip:
Yeah saw the issue with the syntax highlighting. Might be something to think about in the long-term. For now I would say LGTM and iterate. If you want to track the syntax issue, you can open an issue.

PR #52, opened 2018-03-14
Remove leftover data files

### LINKED ISSUE #50: Remove leftover data directory
bug

**What happened**:
After the migration the data directory is not necessary in exampleSite anymore.
 
No data directory in exampleSite.

lauragift21:
Removes all data files after migrating to content structure.

fixes #50


exampleSite/config.toml
```diff
@@ -12,7 +12,7 @@ googleAnalytics = ""
 [params]
   name = "Okkur Labs"
   description = "Open Source Theme for your next project"
-  attribution = "true"
+  attribution = "false"
```
stp-ip: Do we want to make attribution opt-in? Reasoning behind this change?

stp-ip:
Will merge as the other PR was also merged, but for each new feature a new branch with a fresh base is easier to review.

exampleSite/config.toml
lauragift21: I could switch that back to true but add a comment to either enable true or false.

stp-ip: Think it might be best to discuss that. Let's leave it as opt-in for now. Discussion is at #53

PR #64, opened 2018-03-18
Set background on jumbotron hero fragment 

### LINKED ISSUE #37: Background not settable on jumbotron (header fragment)
bug

**What happened**:
Jumbotron defaults to light background.

Background setting in hero fragment used for jumbotron background too.

lauragift21:
No background added to hero fragment enable background on the jumbotron

fixes #37

### DIFF
--- a/layouts/partials/fragments/hero.html
+++ b/layouts/partials/fragments/hero.html
@@ -4,9 +4,13 @@
 
 <header>
   {{ with $hero.header }}
-    <div class="jumbotron text-center header-image mb-0">
+    <div class="jumbotron text-center header-image mb-0
+     {{- printf " bg-%s" $bg -}}
+    ">
   {{ else }}
-    <div class="jumbotron text-center mb-0">
+    <div class="jumbotron text-center mb-0
+     {{- printf " bg-%s" $bg -}}
+    ">
   {{ end }}
   {{ if $hero.particles }}
     <div id="particles-js"></div>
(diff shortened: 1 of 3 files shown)


stp-ip:
Does this change already solve #37? It seems that it would only affect text color for now and not the actual background. Try without an image and a background = dark.

lauragift21:
True without the image and background: "dark" I have a white text and still same actual background

layouts/partials/fragments/hero.html
```diff
@@ -4,9 +4,13 @@
...
   {{ with $hero.header }}
-    <div class="jumbotron text-center header-image mb-0">
+    <div class="jumbotron text-center header-image mb-0
+     {{- printf " bg-%s" $bg -}}
+    ">
   {{ else }}
-    <div class="jumbotron text-center mb-0">
+    <div class="jumbotron text-center mb-0
+     {{- printf " bg-%s" $bg -}}
+    ">
```
stp-ip: Not necessary here, but this could be simplified to:
```
<div class="jumbotron text-center mb-0
  {{- printf " bg-%s" $bg -}}
  {{ with $hero.header }}
    {{ printf " header-image" }}
  {{ end }}
">
```

lauragift21: Alright sure will update that.

stp-ip: It was my code that was unoptimized to beginn with ;)
If you see something similar, you are happy to change it.

PR #25, opened 2018-01-21
Migrate hero fragment to page bundle

lauragift21:
**What this PR does / why we need it**: Migrate data files to page bundles

**Which issue this PR fixes**:
fixes #22 
fixes #13 

**Special notes for your reviewer**: I came up with this migration for the hero fragment. Seems i'm still getting something wrong. i'd like you to review and provide feedback.

**Release note**:

```release-note

```
ToDos:
* [x] Single page content for about|about-split|nested needs to be fixed/tested
* [x] Split layout needs to be fixed/tested
* [x] Changelog with changes in this PR needs to be written (including documenting it's a breaking change)

PR #67, opened 2018-03-19
Move markdown content from <p> to <div>

### LINKED ISSUE #31: Move from `<p>` to `<div>` for anything that could contain markdown content
feature

**What happened**:
Currently `<p>` is being used in some places, which might lead to issues.

Use `<div>` instead of `<p>`.

lauragift21:
Check for all possible occurrences of paragraph tag using markdown content and change to div tag.

fixes #31 

 Already fixed in #25 but this is an overall check.

### DIFF
--- a/layouts/partials/_analytics/piwik.html
+++ b/layouts/partials/_analytics/piwik.html
@@ -23,8 +23,8 @@
   })();
 </script>
 <noscript>
-  <p>
+  <div>
     <img src="{{ .Site.Params.analytics.piwik.url }}/piwik.php?idsite={{ .Site.Params.analytics.piwik.id }}" style="border:0;" alt="" />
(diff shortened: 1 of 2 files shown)


stp-ip:
So that's the only one left? Let me double check and then we can merge and close another issue \o/

stp-ip:
There is one more instances as far as I can see, but doesn't seem it should be a problem, but wanted to get your opinion on it: https://github.com/okkur/syna/blob/5afa06e98e85b7343e0452fcaebf1b94bfc1f34b/layouts/partials/fragments/hero.html#L54

lauragift21:
Yeah, we should totally change this one too. I did an overall check with the input for <p> so I missed this one.

PR #96, opened 2018-04-28
Move recaptcha config to contact fragment frontmatter

### LINKED ISSUE #88: Set recaptcha sitekey via fragment front-matter
feature

**What happened**:
Currently we set recaptcha sitekey globally.

Set sitekey per contact fragment.

Still needs a discussion. Globally simplifies recaptchas, but it bundles site config with a fragment config. We decided to add scripts from fragments within each fragment. We should probably be consistent and also set fragment specific config to their respective front-matter. Aka sitekey.

Most users with one contact form won't notice. For bigger users this duplicates the config in various front-matters, but also enables various recaptcha keys.
Thoughts?

mpourismaiel:
Move recaptcha config from main config file to fragment config

Also made recaptcha optional with one set-back; Submit button is always enabled.

**Which issue this PR fixes** 
fixes #88


exampleSite/content/index/contact.md
```diff
@@ -17,6 +17,10 @@ email = "mail@example.com"
 button = "Send Button" # defaults to theme default
 #netlify = false
 
+# Optional google captcha
+#[params.contact.captcha]
```
stp-ip: Could probably be simplified to just:
```
#[captcha]
#  sitekey = ""
```
No need to have the more complex format.

stp-ip: Or [recaptcha] to make it clear, that it's a setting for google's recaptcha.

static/main.css
```diff
@@ -9691,9 +9691,11 @@ div.form-error ul li {
 
 #particles-js {
```
stp-ip: Separate change so might usually be better as a separate PR.

mpourismaiel:
@stp-ip @lauragift21 I pushed again and removed the previous commit since it was not working properly. I removed the wrong diff in main.css file and fixed a bug in contact.md variable names.

lauragift21:
Looks good to me too. :+1:

static-main/js/form.js
```diff
@@ -13,13 +13,13 @@ if ($.validate) {
     validateOnEvent: true,
     onError: function($form) {
       $('form[id=' + $form.attr('id') +'] .generic-error').removeClass('d-none');
-  },
+    },
```
lauragift21: whitespace included here.

mpourismaiel:
There's a debugger in the code. I don't have access to a laptop write now. I'll fix it in 3-4 hours.

static-main/js/form.js
mpourismaiel: I don't understand.

mpourismaiel: No I added the whitespace myself. Previously the indent level did not match on this line, every other line started at the 3rd level and this line was not indented.

PR #97, opened 2018-04-29
Refactor footer partial to copyright fragment

### LINKED ISSUE #93: Make footer into fragment
feature

**What happened**:
Not all things are fragments yet.

Fragment all the things.

Similar to #89?

mpourismaiel:
Moves footer partial to copyright fragment

fixes #93

### DIFF
--- a/exampleSite/content/index/copyright.md
+++ b/exampleSite/content/index/copyright.md
@@ -1,0 +1,21 @@
++++
+fragment = "copyright"
+#disabled = true
+date = "2016-09-07"
+lastmod = "2017-09-07"
+weight = 1250
+#background = ""
+
+copyright = "" # default: Copyright $Year $params.name
+attribution = true # enable attribution by setting it to true
+
+[[menus]]
+  url = "/about"
+  name = "About Us"
+  weight = 10
+
+[[menus]]
+  url = "#"
+  name = "Another Link"
+  weight = 30
++++
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -12,7 +12,6 @@ googleAnalytics = ""
 [params]
   name = "Okkur Labs"
   description = "Open Source Theme for your next project"
-  attribution = "false" # enable attribution by setting it to true
 
   # Branding options
   [params.branding]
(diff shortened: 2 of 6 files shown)


exampleSite/content/index/copyright.md
```diff
@@ -0,0 +1,21 @@
++++
+fragment = "copyright"
+#disabled = true
+date = "2016-09-07"
+lastmod = "2017-09-07"
+weight = 1250
+#background = ""
+
+copyright = "" # default: Copyright $Year $params.name
```
stp-ip: $param.name is not descriptive here. Probably should be called Site Name or something.

exampleSite/content/index/copyright.md
```diff
@@ -0,0 +1,21 @@
...
+attribution = true # enable attribution by setting it to true
+
+[[menus]]
```
stp-ip: I like that it moved into the fragment. Not sure, if that's the best thing to do in the long run. We are working around the integrated menu support in Hugo. That might be something to bite us once we have multi lingual sites etc. It would also make it harder to use the extended features set for menus. https://gohugo.io/content-management/menus/

Might be ok in the copyright footer, but for the navigation bar that might be different.
But even with it being ok in the footer, we would increase the inconsistencies within the various menus.

Thoughts?

exampleSite/content/index/copyright.md
mpourismaiel: Fixed in 5cb5f94

exampleSite/content/index/copyright.md
mpourismaiel: Fixed in 5cb5f94

stp-ip:
Looks good. Unfortunate, that the menus need to be centrally managed, but we shouldn't reinvent the wheel here. Looping in @lauragift21 for a final review.

stp-ip:
Just did another full review. The background is still using $copyright.backgroundFooter. That wouldn't work with the current fragment config. That needs a fix before merging.

mpourismaiel:
@stp-ip Fixed in 84852225fece8710bd1db78b92e99ab759cc5bc0

PR #101, opened 2018-05-06
Replace jquery form validator with a much smaller version

### LINKED ISSUE #100: Auto disable contact form submit button until validated
bug

**What happened**:
Due to a change the button won't be disabled. That could lead to user problems.

Disabled button until recaptcha and form validation work.

mpourismaiel:
jQuery form validator is bulky and we are using just a fracture of it. This PR uses a small library that I've written (may have bugs I have not seen) and also fixes the issue with the form button not being disabled when captcha is not available.

fixes #100 

This is removing jquery form validator and replacing it with another library. I've tested the contact form as much as I could and every aspect I could think of. But please test it again.


stp-ip:
For this I would love to wait for @lauragift21's review.

On my side. I dislike writing a new library as that always gives a new thing to maintain etc. But I like the approach from the library to focus on regex and make it possible to give global regex options for standard things. Also removing jquery is definitely a good choice.

Overall awesome.

Issue #99, opened 2018-04-30
Github icon not working within star button (navbar)


stp-ip:
**Is this a BUG REPORT or FEATURE REQUEST?**:
bug

**What happened**:
Currently the Github icon doesn't work on the demo.

stp-ip:
For me the usual Github icon doesn't work. It's showing a !? instead. Didn't debug it yet. Just wanted to file the bug already before I forget ^^

stp-ip:
My assumption would be that, while we automated the font library, we also updated it and the convention on how to call brand icons changed somehow or we need to especially reference brand icons in the automation.

lauragift21:
Oh, right I see what you mean.

PR #102, opened 2018-05-06
Replace fontawesome svg icons with fonticons

mpourismaiel:
Removing fontawesome svgs would remove 900kb (not minified, not gzipped) and does not seem to impact the theme in any other way.


stp-ip:
We can go with this solution for now. SVG was choosen as there might be an option to lazy load svgs or reduce the number embed to the ones used.
Happy to go with the more performant solution first.

stp-ip:
Sidenote: This doesn't fix #99 yet.

PR #103, opened 2018-05-06
Change variable names in config.toml to snake case

### LINKED ISSUE #87: Migrate config.toml to snake_case
feature

**What happened**:
We decided to use snake_case per recommendation from Hugo.

Use snake_case in config.toml.

Even so Hugo doesn't support camelCase well and recommends snake_case in config.toml they actually use camelCase :man_facepalming: At least we should be consistent and also use camel_case in config.toml.

mpourismaiel:
Change config.toml to snake_case

fixes #87


stp-ip:
Let's do the fix, after we merge #108 and with that the update to Hugo 0.40.3.

mpourismaiel:
@stp-ip Is there any convention Hugo keeps true? googleAnalytics is lower camelCase and LanguageCode is upper camelCase while Hugo claims snake_case is preferred. Anyways, could you please check it again?

stp-ip:
Seems like in that regard hugo is horrible. Seems good will test it and merge.

exampleSite/config.toml
```diff
@@ -1,6 +1,6 @@
 # Site settings
 baseurl = "/"
-languageCode = "en-us"
+LanguageCode = "en-us"
```
stp-ip: It can be lowercase here, but needs to be uppercase in the html layouts afaik.

exampleSite/config.toml
```diff
@@ -15,26 +15,26 @@ googleAnalytics = ""
...
-    textNavigation = "Syna"
+    logo_navigation = "logo.svg"
+    text_navigation = "Syna"
 
   [params.style]
     #background = "secondary"
-    #backgroundNavigation = "light" # white, light, secondary, primary, dark - default: dark
-    #backgroundFooter = "light" # white, light, secondary, primary, dark - default: dark
+    #background_navigation = "light" # white, light, secondary, primary, dark - default: dark
+    #background_footer = "light" # white, light, secondary, primary, dark - default: dark
```
stp-ip: background_footer isn't used anymore correct? As we moved the copyright into it a fragment.

layouts/partials/fragments/copyright.html
```diff
@@ -48,11 +48,11 @@
           </div>
         </div>
       {{ end }}
-      {{ if $menus.copyright_footer }}
+      {{ if .Site.Menus.legal_footer }}
```
stp-ip: We are not quite consistent with the naming. The fragment is copyright. The translation key is legalFooter and the menu is legal_footer. Probably should be all moved to copyright or something more abstract.

exampleSite/config.toml
mpourismaiel: Fixed in 02991898802d2faa7c8be867619a655643a2edb9

exampleSite/config.toml
mpourismaiel: Fixed in 02991898802d2faa7c8be867619a655643a2edb9

layouts/partials/fragments/copyright.html
mpourismaiel: Fixed in 02991898802d2faa7c8be867619a655643a2edb9

exampleSite/config.toml
mpourismaiel: Fixed in 02991898802d2faa7c8be867619a655643a2edb9

PR #108, opened 2018-05-11
Add support for global fragments

### LINKED ISSUE #92: Introduce global fragments
feature

**What happened**:
With the idea of making everything into a fragment the theme becomes really modular and powerful, but it also opens up sites to use a lot of duplication and complexity.

Introduce a way to set global fragments.

How could something like that look like and work?
A few options in my mind:
1. inject/add: global/navbar.md -> injects itself into all pages as additional fragment
2. default: global/navbar.md -> injects itself into all pages, that don't have a specific fragment yet
3. overwrite: global/navbar.md -> overwrites all fragments with the global config

mpourismaiel:
Users now can add more used fragments into global and the theme will automatically use that global fragment if no local fragment is overriding it.

fixes #92


layouts/_default/single.html
```diff
@@ -9,51 +9,32 @@
...
-      {{ partial "fragments/buttons.html" (dict "self" . ) }}
+  {{/*
+    This code may not be understandable immediately. This is trying to find all
+    the resources (read fragments) for each page and all the global fragments.
+    When combining global and local fragments, local ones have more priority.
+    So if there is only one footer fragment and that is global, that footer
+    fragment is shown. The same is true if there are no global footer fragments
+    and one local one. But if there is one global footer fragment and one local
+    footer fragment, the local one is rendered.
+    */}}
```
stp-ip: Thanks for documenting. Would love some thoughts from @lauragift21.
I have the design in my head so it seems fine. I would probably clarify the difference between "fragment" and the custom fragment name aka filename. As the filename would be the factor in overwriting not the "fragment" type/name.

layouts/_default/single.html
```diff
@@ -9,51 +9,32 @@
...
+    {{ if not (where ($.Scratch.Get "fragments") ".Name" .Name) }}
+      {{ $.Scratch.Add "fragments" (slice .) }}
     {{ end }}
+  {{ end }}
 
-    {{ if and (not .Params.disabled) (in .Params.fragment "custom") }}
-      {{ partial (print "fragments/" .Params.fragment_name ".html") (dict "self" .) }}
+  {{ range sort ($.Scratch.Get "fragments") "Params.weight" }}
+    {{ if and (not .Params.disabled) (isset .Params "fragment") (not (eq .Params.fragment "hero")) }}
+      {{ partial (print "fragments/" .Params.fragment ".html") (dict "menus" $.Site.Menus "Site" $.Site "resources" $.Resources "self" .) }}
```
stp-ip: I would have assumed we used $.Scratch in the fragments so that not providing "$" or "tmp_full" would break hero, table and item.

stp-ip: Considering this we might actually just move to "$" and "." in this PR. Thoughts?

layouts/_default/single.html
```diff
@@ -9,51 +9,37 @@
...
+    {{ range ($.Scratch.Get "global_fragments") }}
+      {{ if not (where ($.Scratch.Get "fragments") ".Name" .Name) }}
+        {{ $.Scratch.Add "fragments" (slice .) }}
+      {{ end }}
     {{ end }}
+  {{ end }}
 
-    {{ if and (not .Params.disabled) (in .Params.fragment "custom") }}
-      {{ partial (print "fragments/" .Params.fragment_name ".html") (dict "self" .) }}
+  {{ range sort ($.Scratch.Get "fragments" | default slice) "Params.weight" }}
```
stp-ip: What's the reasoning behind default slice?

layouts/_default/single.html
```diff
@@ -9,51 +9,37 @@
...
+    {{ if and (not .Params.disabled) (isset .Params "fragment") (not (eq .Params.fragment "hero")) }}
+      {{ partial (print "fragments/" .Params.fragment ".html") (dict "menus" $.Site.Menus "Site" $.Site "resources" $.Resources "self" .) }}
```
stp-ip: Fascinating we are using $.Scratch in hero (PR), table and item, but seem not to run into issues. Have to check that out.

layouts/_default/single.html
lauragift21: Personally, I think going with the global fragment method is way better. It simplifies the layout and is a lot more reusable in terms of adding another "fragment", that brings me to ask do we still need to add the split layout using this method or everything is just going to be within a single layout?

lauragift21: Alright. this explains it better.

stp-ip:
Let's fix the broken build and then add the same logic to the split layout. Then we should be able to merge.
I'll be doing some testing tomorrow until we have the above things done.

layouts/_default/single.html
mpourismaiel: $.Scratch is scoped and can be used inside partials with their own scopes so not passing the $ for top level in pages to partials doesn't make $.Scratch unusable since $ in partials refers to their own top level scope.

layouts/_default/single.html
mpourismaiel: Let's continue this one in the other discussion https://github.com/okkur/syna/pull/108#discussion_r187721597

layouts/_default/single.html
mpourismaiel: Sometimes there are simply no fragments. For example nesting page has no local fragments and if you remove global fragments the loop would crash since `$.Scratch.Get "fragments"` would be `nil`. `default slice` seemed cleaner than an if block.

stp-ip: Thanks. Makes sense.

mpourismaiel:
I fixed the conflict and the issue with build (thanks @stp-ip). Also added the same code to `split` layout, although I changed the code to match the previous way split was rendered (weirdly, hero is rendered after content in there which would be fixed if we decide to make content partials fragments as well). I think everything is in order.

stp-ip:
Great. The "broken" split layout will be fixed with having content a fragment. For now it doesn't break the example page and it won't be released without content being a fragment.

/LGTM

PR #114, opened 2018-05-14
Remove jquery and replace it with very smaller class

### LINKED ISSUE #107: Remove unnecessary JS resources (jQuery/BoostrapJS)
Actually a refactor

**What happened**:
Static files that are in the theme are bulky and can be smaller. jQuery is the largest part of the theme right now and is not used very much. It is a dependency of bootstrap but we're not using any part of bootstrap js as far as I can see. It is possible to remove jQuery and bootstrap js and if needed, use BootstrapNative.

The theme to be smaller in size when served across the web.

Open demo website and check the bundle size.

mpourismaiel:
Replaces jquery with an smaller class

fixes #107


stp-ip:
Will do another review tomorrow with a fresh set of eyes. Would love a review from Gift in the interim. Looks good so far. Cuts down unnecessary jQuery usage \o/

PR #118, opened 2018-05-17
Refactor layouts and make content fragments

### LINKED ISSUE #117: Content and content-split fragments
Actually more of a refactor

**What happened**:
Single and split layouts display content and those contents can be fragments. Them not being fragments causes issues such as two layouts with only a single part different, not being able to make nav a fragment with ease and two different loops to load fragments.

Content and content split be two fragments and remove split layout.

mpourismaiel:
Make content fragments and remove split layout

fixes #117


exampleSite/content/about-split/content-split.md
```diff
@@ -0,0 +1,42 @@
++++
+fragment = "content-split"
+#disabled = true
+date = "2016-09-07"
+weight = 100
+#background = ""
+title = "About Syna"
+subtitle = ""
+
+content = """
```
stp-ip: Content is easier to be done within the actual content section. Makes formatting for the user easier. Basically moving this to after `++++` and then calling it via .Content should work.

exampleSite/content/about-split/content-split.md
```diff
@@ -0,0 +1,42 @@
...
+Etiam eget metus elit. Praesent turpis mi, malesuada nec libero a, cursus varius leo. In malesuada gravida est, nec maximus turpis posuere eu. Proin rutrum mattis lectus. Ut mi elit, tincidunt ut aliquet consequat, facilisis vitae arcu. Proin ut massa auctor, imperdiet sem ac, varius justo. Integer porta tellus quis eros sollicitudin mollis. Donec nulla sem, porttitor et velit vitae, tincidunt congue erat. Vivamus urna libero, viverra non pharetra eu, bibendum id arcu. In fermentum bibendum purus, vitae pharetra ipsum gravida quis. Nunc vitae congue diam. Maecenas pharetra orci vel magna fermentum, vitae auctor sapien consectetur. Nunc tempor erat vel nisi ullamcorper mollis. Sed et lectus accumsan, pharetra nunc scelerisque, finibus leo. Sed tempor eros nec nunc porta sodales.
+"""
+
+[sidebar]
+  title = "Sidebar"
+  content = """
+So much information
+Phone numbers
+Details
+Or even more
```
stp-ip: Without 2 trailing whitespaces will render it on a single line. Probably not what we want here.

exampleSite/content/about/content.md
```diff
@@ -0,0 +1,35 @@
...
+#disabled = true
+date = "2017-10-05"
+weight = 100
+#background = ""
+
+title = "About Syna"
+subtitle = ""
+
+
+content="""
```
stp-ip: Same as with split.

exampleSite/content/nesting/content.md
```diff
@@ -0,0 +1,39 @@
++++
+fragment = "content-single"
+#disabled = true
+date = "2017-10-05"
+weight = 100
+#background = ""
+description = "Nested Page"
+
+content = """
```
stp-ip: Same as with split.

exampleSite/content/nesting/content.md
```diff
@@ -0,0 +1,39 @@
++++
+fragment = "content-single"
+#disabled = true
+date = "2017-10-05"
+weight = 100
+#background = ""
+description = "Nested Page"
```
stp-ip: Probably should be `title`?

layouts/partials/fragments/hero.html
```diff
@@ -1,9 +1,8 @@
 {{ $hero := .self.Params }}
-{{ $dot := .dot }}
 {{ $name := .self.Name }}
 {{ $name := replace $name "/index.md" "" }}
 {{ $name := strings.TrimSuffix ".md" $name }}
-{{ $path := $dot.File.Path }}
+{{ $path := .self.File.Path }}
```
stp-ip: Haha totally missed that as a workable solution.

exampleSite/content/about-split/content-split.md
stp-ip: If that doesn't work, it would make the whole interface for the user a pain for content.

mpourismaiel: Actually tried that solution and it didn't work. Entire resource scope did not contain the markdown that comes after frontmatter. No idea why that happens though. And I agree this is ridiculous but I have no idea how to ease the pain.

exampleSite/content/nesting/content.md
mpourismaiel: It's the mirror of https://github.com/okkur/syna/pull/118/files#diff-fadb7b097ab29fdd0268523ca6c0e097L5

mpourismaiel: Should probably add title though.

layouts/partials/fragments/hero.html
mpourismaiel: Yeah I was surprised to see $dot in the single layout.

exampleSite/content/about-split/content-split.md
stp-ip: Damn will look into it.

exampleSite/content/nesting/content.md
stp-ip: Yeah, but I think that was a leftover mistake I would assume. We don't use description at all so it probably is save to migrate that to title.

stp-ip:
One thing that might be annoying for users is the need for a content file to duplicate the title.
True it is more explicit, but is a bit annoying. We can leave it for now and if it becomes a user issue later, try to find a workaround. Don't think there is an easy one so.

exampleSite/content/about-split/content-split.md
```diff
@@ -4,10 +4,17 @@ fragment = "content-split"
...
 
-content = """
+[sidebar]
+  title = "Sidebar"
+  content = """
+So much information
+Phone numbers
+Details
+Or even more
+"""
```
stp-ip: Trailing whitespace ;)

exampleSite/content/about/content.md
```diff
@@ -4,12 +4,8 @@ fragment = "content-single"
 date = "2017-10-05"
 weight = 100
 #background = ""
-
-title = "About Syna"
-subtitle = ""
-
-
-content="""
+subtitle = "asdfasdfsaf"
```
stp-ip: Not sure I agree with the choice for example data here ^^
Also title could be added or what was the reason to scrape it?

exampleSite/content/nesting/content.md
```diff
@@ -4,9 +4,7 @@ fragment = "content-single"
 date = "2017-10-05"
 weight = 100
 #background = ""
-description = "Nested Page"
-
-content = """
```
stp-ip: Let's add commented and empty titles, just as easy of use for people setting something up:
```
#title = ""
#subtitle = ""
```

exampleSite/content/about/content.md
mpourismaiel: LMFAO That was not supposed to be seen.
I'll readd title.

exampleSite/content/nesting/content.md
mpourismaiel: I removed `# Nested page` and added it as title. Also commented all the subtitles.

stp-ip:
/LGTM
After fixing the conflicts that is. @lauragift21 please review again. Might merge before an approve from your side as it's the base for a few other issues, but if you find any additional issues we can do a follow on PR.

PR #119, opened 2018-05-17
Make navbar a fragment

### LINKED ISSUE #89: Make navigation a fragment
feature

**What happened**:
Currently the navigation is "hardcoded" and hidden, when empty.

Migrate navigation to its own fragment.

Navigation fragment would enable stacked navigation and per page navigation.

Needs a bit of thinking.
How to use with auto hide feature?
Where should the menu config be set?
How does it work within the current navigation config from Hugo?
Set menus in config.toml, but set used menu in each fragment?
Too much complexity for multi page sites, that just need one navbar? <-- workaround with a global navbar fragment?

mpourismaiel:
Makes navbar a fragment instead of a hardcoded partial.

fixes #89 

This is based on #118 so please wait for that to be merged.

**Release note**:
```release-note
Navbar is now a fragment
```

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -44,6 +44,34 @@ googleAnalytics = ""
   [params.fontawesome]
     disabled = false
 
+# main/prepend/postpend menus are displayed in nav fragment.
+# Keep an eye out for updates. These will be moved out to make navbar more
+# configurable.
+[[menu.main]]
+  url = "#"
+  name = "Long Link"
+  weight = 10
+
+[[menu.main]]
+  url = "/about-split"
+  name = "Page-split"
+  weight = 30
+
+[[menu.main]]
+  url = "/about"
+  name = "Page-single"
+  weight = 30
+
+[[menu.main]]
+  url = "#contact-fragment"
+  name = "Contact"
+  weight = 50
+
+[[menu.main]]
+  url = "#"
+  name = "Link"
+  weight = 30
+
 [[menu.footer]]
   url = "#"
   name = "Link"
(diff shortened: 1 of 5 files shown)


layouts/partials/fragments/nav.html
```diff
@@ -1,3 +1,7 @@
+{{ $nav := .self.Params }}
+{{ "<!-- Nav -->" | safeHTML }}
+{{ $bg := $nav.background | default "secondary" }}
+
 <nav class="overlay navbar navbar-expand-lg py-2
   {{- $bg := .Site.Params.style.background_navigation | default "dark" -}}
```
stp-ip: Seems left over.

mpourismaiel:
@stp-ip Navbar menus are now alongside other menus in `config.toml`. I don't think Hugo would do anything regarding menus in resources and there is now way around it. If I figure out a way to do something like this (this is done in JS, don't now the equivalent in Go or Go templates) it might be less of a pain.

```javascript
.Site.Menus[`navbar-${.self.Params.menu_name}-prepend`]
.Site.Menus[`navbar-${.self.Params.menu_name}-main`]
.Site.Menus[`navbar-${.self.Params.menu_name}-postpend`]
```

exampleSite/config.toml
```diff
@@ -44,6 +44,34 @@ googleAnalytics = ""
   [params.fontawesome]
     disabled = false
 
+# main/prepend/postpend menus are displayed in nav fragment.
+# Keep an eye out for updates. These will be moved out to make navbar more
```
stp-ip: Great adding a future compatibility change notice here.

stp-ip:
I think we can merge this for now and I'll try to find a Go templates way for the above idea.

Issue #124, opened 2018-05-22
Add resource fallthrough to all images


stp-ip:
**Is this a BUG REPORT or FEATURE REQUEST?**:
feature

**What happened**:
Currently the only image that has a fallthrough mechanism using `fragment > page > global` is hero logo.

**What you expected to happen**:
All images should use that mechanism.

**Anything else we need to know?**:
Might be easier to implement an image shortcode first via #95

PR #141, opened 2018-05-28
Add shortcodes

### LINKED ISSUE #95: Simplify code and helper snippets via shortcodes
feature

**What happened**:
Currently we have quite a bit of duplication in the various fragments.

Reduce duplication and move helper snippets into shortcodes.

https://gohugo.io/templates/shortcode-templates/

mpourismaiel:
Adds semi-shortcode partials that would allow for DRY code and some functions

fixes #95 

This only implements partials for logo and img (with fallthrough support).

### DIFF
--- a/layouts/partials/fragments/footer.html
+++ b/layouts/partials/fragments/footer.html
@@ -19,15 +19,7 @@
           <h4>
             {{- .title -}}
           </h4>
-          {{ if .image }}
-          <a href="{{ .link }}">
-            <img src="{{ (printf "images/%s" .image) | relURL }}" class="img-fluid w-50" alt="{{ .text }}">
-          </a>
-          {{ else }}
-          <div>
-            {{- .text -}}
-          </div>
-          {{ end }}
+          {{ partial "shortcodes/logo.html" (dict "src" ((printf "images/%s" .image) | relURL) "text" .text "url" .link "image_class" "img-fluid w-50") }}
         {{ end }}
 
         {{ if $menus.footer_social }}
--- a/layouts/partials/fragments/hero.html
+++ b/layouts/partials/fragments/hero.html
@@ -24,32 +24,10 @@
   {{ end }}
 
   {{ if $hero.logo }}
-    {{/* fallback to global resource */}}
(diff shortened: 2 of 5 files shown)


mpourismaiel:
@stp-ip Should I add fall through for item and member fragments as well? Or a new partial for loading images.

Also loading partials has one issue. Using `dict` for arguments makes it so that list of variables would make the line maybe reach 200 characters in length. Even using variables to store larger calculations doesn't make it much smaller. Maybe this isn't that good of an idea (using partials) and we should use it just for situations like resource fallthough.

mpourismaiel:
Also the `partial/shortcodes/img.html` would handle resource fallthrough for images and I guess for now there is no other resource that we may want to fallthrough. So maybe with this, #124 would be resolved as well.

stp-ip:
Great for fighting with shortcodes and the partial workaround. The syntax of the "function" call is long and hard to understand. Question is, if that is better then code duplication.

`{{ partial "shortcodes/img.html" image }}` would feel easier, while complicating the preparation code of $image.
$image itself would have the following data embed:
* .image <-- filename to look for
* .text <-- alt name
* .link <-- optional linking
* .class <-- maybe optional class data

Not sure about class itself, but condensing that would be great.

Might have been a bad idea on my part to try deduplicating the fallback code. Would love to hear other thoughts as well. @lauragift21

mpourismaiel:
How should I make a structured data containing those variables? 

Does hugo have any such thing planned? Something like custom functions or shortcodes or even better variable declarations?

stp-ip:
Nothing that comes to mind. Will look through the possibilities tomorrow.

mpourismaiel:
@stp-ip I'm still lost on this. I guess it would be better to push this issue a bit so Hugo may come up with a better solution. Using this style is really horrible.

stp-ip:
Yeah I think duplicating the code is the better choice for now. Ok let's close this one here and go with implementing fallthrough for all images via code duplication instead.

mpourismaiel:
So we're going to push this to 0.12?
-- 
Regards,
Mahdi Pouresmaiel

stp-ip:
Push the refactoring to post v0.11. Probably depends on a good Hugo way to do that. So no set milestone for now.

PR #147, opened 2018-05-29
Remove extra whitespace in layout files

### LINKED ISSUE #60: Reduce whitespace within Go template
feature

**What happened**:
Currently we sometimes cut out whitespace via hugo's `-` operator and mostly we don't.
This leads to non obvious issues with cut out whitespace within html and also to a lot of whitespace in the final html without cutting it out.

More optimized html.

Move to using `{- ... -}` everywhere by default. Fix any issues coming up with that change.

mpourismaiel:
Remove extra new lines in layout files (layouts, partials and fragments)

fixes #60 

I removed as much as was possible and in some places I couldn't figure out how to avoid smelly HTML output. If there is anything too serious let me know. (Also God bless VIM macros)

### DIFF
--- a/exampleSite/layouts/partials/fragments/sample-custom-fragment.html
+++ b/exampleSite/layouts/partials/fragments/sample-custom-fragment.html
@@ -1,7 +1,6 @@
-{{ $fragment := .self.Params }}
-{{ $name := strings.TrimSuffix ".md" (replace .self.Name "/index.md" "") }}
-{{ $bg := $fragment.background | default "secondary" }}
-
+{{- $fragment := .self.Params -}}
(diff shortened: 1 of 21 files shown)


stp-ip:
Vim \o/
It's already simplifying the output quite a bit. Only thing that feels a bit wrong is that it's selectively choosing depending on the various cases. There is no rule for how to do whitespace reducing, which could be used as a guide for PRs etc. That might lead to harder to handle reviews etc.
I assume hard removing all whitespace is going to uglify the html quite much.

Thoughts?

mpourismaiel:
There isn't anything I can think of. I did these changes based on two rules:

1. If the next line contains go template remove left and right side whitespaces, if html only remove left side whitespace.
2. If the go template is in the same line as HTML remove only side of whitespaces that is not printed by go template.

PR #170, opened 2018-06-02
Modify joined to date within member frontmatter

### LINKED ISSUE #164: Member content files doesn't have date, but uses "joined"
feature

**What happened**:
Currently member content files use `joined` instead of `date`.

Member files might use `date` instead of joined.

Joined is more descriptive, but date should be provided anyway. So either we add date, which would duplicate time data or we move `joined` to `date`. Would make it more consistent in my view.

lauragift21:
Change `joined` to `date` within the member content files for consistency.

fixes #164

### DIFF
--- a/exampleSite/content/index/awesome-team/donald.md
+++ b/exampleSite/content/index/awesome-team/donald.md
@@ -1,7 +1,7 @@
 +++
   title = "Tiny Gopher 4"
   weight = 30
-  joined = "2017-10-17"
+  date = "2017-10-17"
   image = "tinygopher.png"
   position = "Gopherineer"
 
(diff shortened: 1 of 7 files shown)


stp-ip:
This also needs the changes within the layout files.

lauragift21:
Yes I checked, but i don't think we are calling joined within the fragment.

stp-ip:
No you are right. I'm the one who was wrong. Thanks for checking. I had in the back of my mind the idea that we sorted by `joined`, but we use `weight` instead.
As we don't use it yet we can merge as is. If we later decide to expose `joined` in the interface, we can decide to use `date` or add a new `joined`

PR #173, opened 2018-06-02
Simplify fragments by using .Params without assignment

stp-ip:
Simplify call to .Params and make all fragment code more consistent.


stp-ip:
tmp_full didn't seem like the most describing one. For now it's not that temporary anymore and root seemed more descriptive.

Happy that you agree with the refactoring \o/

PR #175, opened 2018-06-02
Fix Fontawesome

### LINKED ISSUE #159: Fontawesome icons working only partly
bug

**What happened**:
Fontawesome icons not working reliably due to the split into `fab`, `fas`, `far`.

All fontawesome free icons working.

Set `fa-external-link-square-alt` as icon within the member fragment.

Simplest solution seems to force the user to set the whole icon classes `fab fa-gitlab` instead of just `fa-gitlab` and remove the hardcoded variant in the code.

stp-ip:
Enable full range of Fontawesome icons.

fixes #159

Builds on #174 

**Release note**:

```release-note
Fontawesome icons now need to declare the full icon class: `fab fa-facebook` instead of `fa-facebook`
```


stp-ip:
@mpourismaiel merge conflict prevention. We would have run into one with that change. So I based it on #174.

PR #153, opened 2018-05-31
Add resource fallthrough to all fragments with image

mpourismaiel:
**What this PR does / why we need it**:
Add resource fallthough to member and item fragment and fix an issue with it in hero fragment

**Which issue this PR fixes**:
fixes #124 

**Special notes for your reviewer**:
The fix is that the `fileExists` function does not check exactly where the file is. With this fix it looks in the `content` directory for the file to see if the file is available in page or in the fragment directory.

**Release note**:
```release-note
Add resource fallthrough (Hugo style file matching, you can now choose `image.png` and that file is checked in fragment directory, then in page directory and finally in `static/images` directory for that file.)
```

PR #178, opened 2018-06-02
Image fallthrough continuation

stp-ip:
Build on @mpourismaiel 's work and add fallthrough image support to all current images.

fixes #153 

Made the snippet work for all fragments the same. It has 2 parts. One path preparation and one actual src generation, which can either be at the top of the file for single uses, inlined right before the image is used or within a loop.
Also added comments to explain the duplication and a do not edit note.

Sidenote: This currently does not separate file sources into subdirectories in `static/`. If this is the wrong approach, we should probably add a follow-on PR. This one is already huge. (sorry for that)

**Release note**:
```release-note
Add image fallthrough. Images first look in its fragment directory, then within the page directory and last the global "static/" directory.
```


stp-ip:
I move the examples into a dev subdirectory with all fragments and appearances tested. It looks a bit weird for the hero background as I reused the logo SVG, but it should convey the point.

Fragment: /dev/fallthrough/fragment/
Page: /dev/fallthrough/page/
Global: /dev/fallthrough/global/

As always. Please don't hesitate to discuss concerns with any of the approaches or call out mistakes I've made.

stp-ip:
This doesn't have the fix #174 yet. Therefore the member fragment should have a weird second member in /dev/fallthrough/fragment/.

layouts/partials/fragments/footer.html
```diff
@@ -1,10 +1,43 @@
...
+{{- $path := strings.TrimSuffix ".md" $path -}}
+{{- $page_path := replace $path (printf "/%s" .Name) "" -}}
+{{/* End of do not change */}}
+
+{{/* Global resource fallback - can also be used within loops */}}
+{{- $image := .Params.logo.image -}}
+{{/* Do not change the following snippet */}}
+{{/* Code is duplicated throughout the code */}}
+{{- $.root.Scratch.Set "image" (printf "images/%s" $image) -}}
+{{- range $.resources.ByType "image" -}}
```
mpourismaiel: I think this loop is not achieving anything. It's adding an scope and running the code block as many times as there are resources in the page but it's achieving the results of the first loop just several times.

layouts/partials/fragments/footer.html
```diff
@@ -1,10 +1,43 @@
...
 {{- $bg := .Params.background | default "dark" }}
 
+{{/* Resource fallthrough path preparation */}}
+{{/* Do not change the following snippet */}}
+{{/* Code is duplicated throughout the code */}}
+{{- $path := .self.File.Path -}}
+{{- $path := replace $path "/index.md" "" -}}
+{{- $path := strings.TrimSuffix ".md" $path -}}
+{{- $page_path := replace $path (printf "/%s" .Name) "" -}}
+{{/* End of do not change */}}
```
mpourismaiel: These two variables can be passes from the `single.html` layout to the fragments. There we have access to `.self` and `.Name` so there shouldn't be any problem and the code would not be duplicated then.

stp-ip: Good point. Thanks for pointing out.

layouts/partials/fragments/footer.html
stp-ip: Not sure I get the whole point. So you would just remove it?

mpourismaiel: I've read the code several times and can't figure out how it's useful. Am I missing something?

stp-ip: Nope I missed something. ^^

stp-ip:
Will work on the conflicts in a bit. Moved and simplified the fallthrough stuff.

layouts/_default/single.html
```diff
@@ -31,8 +31,15 @@
...
   {{- range sort ($.Scratch.Get "fragments" | default slice) "Params.weight" -}}
     {{- if and (not .Params.disabled) (isset .Params "fragment") -}}
+      {{/* Cleanup .Name to be more useful within fragments */}}
       {{- $name := strings.TrimSuffix ".md" (replace .Name "/index.md" "") -}}
-      {{- partial (print "fragments/" .Params.fragment ".html") (dict "root" $ "Site" $.Site "resources" $.Resources "self" . "Params" .Params "Name" $name) -}}
+
+      {{/* Resource fallthrough path preparation */}}
+      {{- $file_path := strings.TrimSuffix ".md" (replace .File.Path "/index.md" "") -}}
+      {{- $page_path := replace $file_path (printf "/%s" $name) "" -}}
+      {{ $fallthrough := (dict "file_path" $file_path "page_path" $page_path ) }}
```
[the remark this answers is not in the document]
stp-ip: fixed

PR #184, opened 2018-06-04
Enable alert prompt for fragment with content not available

### LINKED ISSUE #30: Alert user where to store members/items etc. for specific fragments, when no data available
 feature

**What happened**: Content files won't be visible except the naming structure is consistent. (i.e `content/index/my-awesome-team.md/` and `content/index/awesome-team.md/`  doesn't match)

**What you expected to happen**: To check if this is working correctly print out the resulting `{{$member}}` if it's empty then print out the location of the files in the command line `{{ printf "%s/*" $name_t }}`

lauragift21:
Add an alert prompt to fragment with content data not available. 

Optional one line note for this specific change, that can be used in a release-note or changelog.

```release-note
Add an alert prompt to fragment with content data not available. 
```

### DIFF
--- a/layouts/partials/fragments/items.html
+++ b/layouts/partials/fragments/items.html
@@ -38,55 +38,54 @@
         </div>
       {{- end }}
       <div class="row justify-content-center align-items-start">
-        {{- if eq $items 0 -}}
+        {{- if eq (len $items) 0 -}}
         <div class="alert alert-danger">
             {{- printf "Currently there is no configured items content files present. To add new items add a content file within %s"  $.fallthrough.file_path -}}
         </div>
         {{- else -}}
-        {{- range $items -}}
-          {{/* Handle special case of index.md being considered an item */}}
-          {{- if not (in .Name "/index.md") -}}
-            {{- $item := .Params }}
-            <div class="col-md-4 text-center">
-              <span class="fa-stack fa-3x m-2">
-                <i class="fas fa-circle fa-stack-2x
-                  {{- if eq $bg "primary" -}}
-                    {{- printf " text-%s" "dark" -}}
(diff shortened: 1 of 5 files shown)


layouts/partials/fragments/member.html
```diff
@@ -37,6 +38,11 @@ <h5>
           </div>
         </div>
       {{ end }}
+      {{- if eq $members 0 -}}
+        <div class="text-danger alert-danger text-center">
+          {{- printf "Currently there is no configured member content files present. To add new members add a content file within index/%s" $path -}}
```
mpourismaiel: Please change the `index/%s` to something that would be correct in other pages as well.

layouts/partials/fragments/items.html
```diff
@@ -38,6 +39,11 @@ <h5>
         </div>
       {{- end }}
       <div class="row justify-content-center align-items-start">
+        {{- if eq $items 0 -}}
+        <div class="text-danger alert-danger">
+            {{- printf "Currently there is no configured items content files present. To add new items add a content file within index/%s" $path -}}
```
mpourismaiel: Change the index here as well.

layouts/partials/fragments/items.html
```diff
@@ -1,4 +1,5 @@
 {{- $items := .resources.Match (printf "%s/*" .Name) -}}
+{{- $path := .Name -}}
```
mpourismaiel: Isn't there any way to avoid doing this? It's a bit annoying.

lauragift21: hehe yeah i didn't want to bring that back seeing it was already removed but i needed a way to get the filename

PR #204, opened 2018-06-26
Restructure single member mode

stp-ip:
Production ready use for single member mode.

https://deploy-preview-204--syna-demo.netlify.com/#single-member-fragment

**Release note**:

```release-note
Improve structure and design of single member mode for member fragment
```

### DIFF
--- a/layouts/partials/fragments/member.html
+++ b/layouts/partials/fragments/member.html
@@ -150,8 +150,8 @@
           {{- end }}
         </div>
       {{- else if le $members 1 -}}
-        <div class="card p-2 m-2">
-          <div class="row justify-content-start align-items-start">
+        <div class="card p-2">
+          <div class="row m-0 justify-content-center align-items-center">
             {{- range $members -}}
               {{/* Handle special case of index.md being considered a member */}}
               {{- if not (in .Name "/index.md") -}}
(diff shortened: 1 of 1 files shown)


layouts/partials/fragments/member.html
```diff
@@ -234,30 +237,32 @@ <h6 class="font-italic mb-0">
...
-                          </a>
-                        </span>
-                      {{- end -}}
-                    </div>
-                    {{- end -}}
-                    {{- with .Content -}}
-                      <div class="pt-3">
-                        <div class="description">
-                          {{- . | markdownify -}}
+                      <div class="col-12 text-center">
```
mpourismaiel: This part of the template is only needed if there are member icons. If there are none then there is a padding introduced by `py-2` on the next element and some weird styles that make `col-12`. IMO only render this part if there are `$member.icons`

stp-ip: Just went through the code. Isn't that already the case?
This is around the mentioned div `{{- if $member.icons -}}.

mpourismaiel: Oh yeah you're right. Sorry. I mistook line 251 as `added` instead of `removed`. Monitor color settings :D

PR #214, opened 2018-06-27
Add editor fragment

### LINKED ISSUE #11: Create editor fragment
feature

**What happened**:
No editor fragment.

Simple editor fragment.

Look for editor fragment.

Note:
Simple editor PoC. No WYSIWYG yet.

mpourismaiel:
Add a new editor fragment that can create any form using a JSON data file.

fixes #11 

This is a POC and needs cleaning up even if it looks good.

**Release note**:
```release-note
Add new fragment: **Editor**
```


stp-ip:
Looks simple and powerful. Making it possible for other tools to generate a json file for the editor.

It's a specific json format so something as the reposeed.json wouldn't work to generate an editor interface:
```
{
  "project": {
    "name": "Reposeed",
    "oneLiner": "Extensive repository base files up and running in an instant",
    "state": "beta",
    "image": "https://github.com/okkur/reposeed/blob/master/media/logo.svg",
    "website": "",
    "description": "Start with the base layer necessary to focus on your project and not on the repository.  \nLicensing, structure, documentation and more boilerplate to get you started from your first commit.\n",
    "version": "v0.4.0"
  },
  "vision": {
    "type": "community",
    "overview": "Providing a generic base layer for new open source repositories.\n",
    "items": [
      "Default usage on first initialization",
      "Generic seed/base layer files",
      "Additions to reposeed only accepted for generic usage",
      "Specific additions should be added to the files after generation"
    ],
    "concept": "Reposeed's core concept is to provide a basic selection of useful templates. With the help of a configuration file and a small substitution engine the project generates a seed layer for new open source repositories.",
    "aim": "Our aim is to provide a sufficient base to built a new open source project on top of. Reducing the burden for maintainers and decreasing the hurdles to get started."
  },
  "contributionLinks": {
    "issueTemplate": "/.github/ISSUE_TEMPLATE.md",
    "starterIssues": "https://github.com/okkur/reposeed/labels/starter%20issue"
  },
  "supportLinks": {
    "documentation": "/docs"
  },
  "readme": {
    "usageExample": "```\ngo get -v -u github.com/okkur/reposeed/cmd/reposeed/...\nmkdir YOUR-PROJECT\n```\n\nRun the following command to create the sample .seed-config.yaml file.\n```\nreposeed init YOUR-PROJECT\n```\n\nChange .seed-config.yaml to your needs.\n```\nreposeed --conf=YOUR-PROJECT/.seed-config.yaml --output=YOUR-PROJECT\n```\n"
  },
  "docs": {
    "development": "Run the following commands to install reposeed and change your directory to reposeed's source code directory.\n```\ngo get -v -u github.com/okkur/reposeed/cmd/reposeed/...\ncd $GOPATH/src/github.com/okkur/reposeed\n```\n\nThen run the following command to install **packr**\n```\nmake packr\n```\n\nThen in case you made any changes on templates, run ```make``` or ```packr install ./cmd/reposeed``` to bundle templates into the generated binary.\n"
  },
  "emails": {
    "commercialSupport": "support@okkur.io",
    "security": "security@okkur.io",
    "coc": "coc@okkur.io"
  },
  "copyright": {
    "owner": "The RepoSeed authors",
    "year": "2017"
  },
  "cla": {
    "copyrightHolder": "Michael Grosser"
  },
  "maintainers": [
    {
      "name": "Michael Grosser",
      "nick": "stp-ip"
    }
  ],
  "repo": {
    "link": "https://github.com/okkur/reposeed",
    "type": "Github"
  },
  "badges": [
    {
      "image": "https://img.shields.io/badge/state-beta-blue.svg",
      "alt": "state"
    },
    {
      "image": "https://img.shields.io/github/release/okkur/reposeed.svg",
      "link": "https://github.com/okkur/reposeed/releases",
      "alt": "release"
    },
    {
      "image": "https://img.shields.io/github/license/okkur/reposeed.svg",
      "link": "LICENSE",
      "alt": "license"
    }
  ],
  "reposeed": {
    "configVersion": "v1"
  }
}
```
A few questions to clarify the context:
* Should we take an arbitrary "output" json file and generate the input editor on top (see above)? That would make it more complex I assume.
* Do we need the json data input or might it be more user accessible to be a content file structure, which can also be JSON in the frontmatter.
* What are the use cases for the current implementation?

Would also love to hear @lauragift21 's thoughts

mpourismaiel:
So the current implementation can be used to implement any form the user requires in the given layout. With this and an added support for a two column layout and captcha support it can be used instead of the contact fragment to give more flexibility and power to the user.

As for the reposeed and resumic, it's not possible to generate the form without JS since the form may change in the future and any change to the form needs a change in the data file if we're going to use this fragment. 

I can add support for a more dynamic data structure (with JS it would be much easier) but with the JSON reposeed presents it's not that possible. For each field we need to know what kind of input should be shown and for accessibility's sake we need labels and validations and other stuff as well.

stp-ip:
> So the current implementation can be used to implement any form the user requires in the given layout. With this and an added support for a two column layout and captcha support it can be used instead of the contact fragment to give more flexibility and power to the user.

Yeah it felt more related to contact than the rest. From a simplicity perspective I think it might still be worth it to keep contact separated, but not opposed to moving in a more general direction.

> As for the reposeed and resumic, it's not possible to generate the form without JS since the form may change in the future and any change to the form needs a change in the data file if we're going to use this fragment.

Having the user change the data file when an API changes should be a prerequisite. So it doesn't have to be completely dynamic.

> I can add support for a more dynamic data structure (with JS it would be much easier) but with the JSON reposeed presents it's not that possible. For each field we need to know what kind of input should be shown and for accessibility's sake we need labels and validations and other stuff as well.

Ok I see the point. But considering that using a payload JSON file wouldn't work, we might be better off using a content file structure instead of a data file I assume. The user needs to configure it anyway so we would assume they can do that within the content structure. The frontmatter can be done in various formats including JSON.

Seems my idea was a bit too overambitious or possible, but mostly not as usable due to missing most of the necessary meta data. Would a JSON Schema file have enough metadata? It has types etc.

> IMO opinion we should make this fragment better and work on the reposeed form data to make it easier to implement it with a JS code that works like this fragment but in a more dynamic manner.

I mostly agree. Let's split it up into various decisions:
* Data vs content file for configuration?
* Contact fragment separate from editor fragment?
* Specific dynamic use case? Load the file from a server and do build the fragment client side? I think that's not necessary.
* JSON Schema enough metadata to use as base for editor?
* Current implementation with full config aka based on reposeed JSON structure enough as a PoC?
* Reposeed editor design consideration. Plain editor fragment enough or should we "fork" it and add as a custom fragment.
* -> Which would lead to the question is the editor framework too plain to be useful?

mpourismaiel:
Okay so I tried some solutions in order to find the answers.

- Content file is a much better solution,

mpourismaiel:
Okay so I tried some solutions in order to find the answers.

> Data vs content file for configuration?
Content file is a much better solution, but that requires a second level to contain the data since the first level of the JSON content file would be used for Hugo itself and the fragment declaration part. Does that sound good?

> Contact fragment separate from editor fragment?

I was thinking of removing the contact fragment altogether and use a sample of the editor fragment to create the contact form. It would be much more flexible but sacrificing the simplicity. Although one can argue that none of these two approaches are simple but at least the editor fragment is much more flexible.

> JSON Schema enough metadata to use as base for editor?

I can't see any shortcomings. It would be a complex file for complex forms but that's needed. Unless we devise some defaults in which case we need a more in depth discussion.

> Current implementation with full config aka based on reposeed JSON structure enough as a PoC?

> Reposeed editor design consideration. Plain editor fragment enough or should we "fork" it and add as a custom fragment.

Implementing an editor for reposeed needs some changes to it's JSON schema since the current implementation is not easy to decipher. It's easy to understand what each part requires when you look at it but implementing it using a general fragment like the current editor fragment needs a lot of changes to the schema. There are three options to implement reposeed editor. Change the schema to a more standardized version, implement it using JS on runtime or create a custom fragment. I'd rather an standard schema.

Considering this fragment is implemented with the general user in mind, it's complicated but flexible and can be used to implement almost any form if we add support for check boxes, radio options, select boxes and file inputs. If we do those inputs then it's possible to use the same fragment for reposeed and resumic editors considering their JSON schema is standardized.

stp-ip:
> Okay so I tried some solutions in order to find the answers.
> 
>     Data vs content file for configuration?
> 
> Content file is a much better solution, but that requires a second level to contain the data since the first level of the JSON content file would be used for Hugo itself and the fragment declaration part. Does that sound good?

Second level seems hard for non json I assume, but having a separate single file similar to the team and items fragments seems worse. Let's go with a second level and use JSON as frontmatter to make the second level easier and let us focus on JSON for now.

> 
>     Contact fragment separate from editor fragment?
> 
> I was thinking of removing the contact fragment altogether and use a sample of the editor fragment to create the contact form. It would be much more flexible but sacrificing the simplicity. Although one can argue that none of these two approaches are simple but at least the editor fragment is much more flexible.

That would add netlify code to the editor for example.
How about using the same base code, but keeping a "fork" of the editor fragment with netlify changes etc. as contact form.
@lauragift21 What's your take on this?
> 
>     JSON Schema enough metadata to use as base for editor?
> 
> I can't see any shortcomings. It would be a complex file for complex forms but that's needed. Unless we devise some defaults in which case we need a more in depth discussion.

So what I mean was. Currently reposeed.json is just key value pairs. We could construct a json schema from that, which would add information such as type etc. This could then be used as base for the editor framework instead of creating a new JSON file just for the editor. But maybe creating a new JSON is the simpler choice and more in line with what people would be using.
@lauragift21 thoughts?

> 
>     Current implementation with full config aka based on reposeed JSON structure enough as a PoC?
> 
>     Reposeed editor design consideration. Plain editor fragment enough or should we "fork" it and add as a custom fragment.
> 
> Implementing an editor for reposeed needs some changes to it's JSON schema since the current implementation is not easy to decipher. It's easy to understand what each part requires when you look at it but implementing it using a general fragment like the current editor fragment needs a lot of changes to the schema. There are three options to implement reposeed editor. Change the schema to a more standardized version, implement it using JS on runtime or create a custom fragment. I'd rather an standard schema.

What specific changes would you do? Happy to make the reposeed config cleaner anyway.
So what would a standardized version look like?
Custom fragment might be something in the long run. Start simple and then go from there. Especially with resumic we most likely need a separate fragment later on. It's just too complex for a simple editor fragment abstraction I think.

> 
> Considering this fragment is implemented with the general user in mind, it's complicated but flexible and can be used to implement almost any form if we add support for check boxes, radio options, select boxes and file inputs. If we do those inputs then it's possible to use the same fragment for reposeed and resumic editors considering their JSON schema is standardized.

As a PoC I agree. Later on people might add their own customized fragment to give their editor a better design tailored to their content.

mpourismaiel:
For now let's finalize the form JSON schema and make it standard. After that Resumic and Reposeed should pass their Schema in the same standardized format.

Also we need file input, select, checkbox and radiobox. Should I go with bootstrap designs?

PR #219, opened 2018-07-06
Add category support for content-single and content-split fragments

mpourismaiel:
Add category support for content-single and content-split fragments

**Release note**:
```release-note
Add category support to content (single and split) fragments
```


stp-ip:
Clarification:
about-split is adjusted to the background color it seems, but about isn't.

mpourismaiel:
That's because someone didn't save the file after they changed the color switch. Also the color switch is based on whether the background color is dark or not. Should I do something else?

Will add a page and push in a moment.

stp-ip:
Nope dark or not should be enough for now. Similar to how we do it with the other switches.

stp-ip:
LGTM

PR #220, opened 2018-07-08
Add blog support

### LINKED ISSUE #4: Create blog layout
feature

**What happened**:
No blog page/post layout is available.

Support for a blog/post layout.

Try to run a blog with syna.

**Environment**:
- Syna Theme version:  v0.7.0
- Hugo version:  v0.30.2

mpourismaiel:
Add simple blog support. This PR adds support for index files to be fragments themselves to make smaller pages easier to create. Also adds list fragment that is able to list pages from different sections and is used in the exampleSite to list latest articles.

fixes #4 

**Release note**:
```release-note
- Add new list fragment that lists pages from a single section
- Index files now can be fragments
```

### DIFF
--- a/exampleSite/content/blog/_index/index.md
+++ b/exampleSite/content/blog/_index/index.md
@@ -2,6 +2,7 @@
 date = "2018-07-06"
 fragment = "list"
 weight = 100
+url = "/blog"
 
 #count = 10 # Default value is 10
 #section = "" # Default value is current section
--- a/exampleSite/content/blog/_index/index.md
+++ b/exampleSite/content/blog/_index/index.md
@@ -1,0 +1,6 @@
++++
+date = "2018-07-06"
+url = "/blog"
+fragment = "list"
+weight = 100
++++
--- a/layouts/partials/fragments/list.html
+++ b/layouts/partials/fragments/list.html
@@ -3,7 +3,7 @@
   where functions achieving this. First where gets all the pages in blog
   section, the second one removes current page from the mix and the last one
   removes the section (.Site.Pages returns the current section as a page). */}}
-{{- $articles := where (where .Site.Pages "Section" (.Params.section | default .page.Section)) "Params.fragment" "in" (slice "content-single" "content-split") -}}
+{{- $pages := where (where .Site.Pages "Section" (.Params.section | default .page.Section)) "Params.fragment" "in" (slice "content-single" "content-split") -}}
 {{- $bg := .Params.background | default "light" }}
 
 {{ "<!-- Blog -->" | safeHTML }}
@@ -18,7 +18,7 @@
             {{- printf " text-muted text-%s" "secondary" -}}
           {{- end -}}
         ">
-          {{- range first (.Params.article_count | default 10) $articles }}
+          {{- range first (.Params.count | default 10) $pages }}
             <article class="col-12 mb-4">
               {{- if .Params.title }}
                 <div class="col-12 pl-0
--- a/layouts/partials/fragments/list.html
+++ b/layouts/partials/fragments/list.html
@@ -1,0 +1,108 @@
+{{/*
+  This variable stores all the page bundles in the blog section. There are three
+  where functions achieving this. First where gets all the pages in blog
+  section, the second one removes current page from the mix and the last one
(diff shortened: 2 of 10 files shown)


layouts/partials/fragments/list.html
```diff
@@ -0,0 +1,108 @@
+{{/*
+  This variable stores all the page bundles in the blog section. There are three
+  where functions achieving this. First where gets all the pages in blog
+  section, the second one removes current page from the mix and the last one
+  removes the section (.Site.Pages returns the current section as a page). */}}
+{{- $articles := where (where .Site.Pages "Section" (.Params.section | default .page.Section)) "Params.fragment" "in" (slice "content-single" "content-split") -}}
```
stp-ip: That's is quite a great query. Wouldn't have thought of that.
It implicitly defaults to .page.Section. Does that always work? Or might there be some strange edge cases?

exampleSite/content/blog/_index/index.md
```diff
@@ -0,0 +1,6 @@
++++
+date = "2018-07-06"
+url = "/blog"
```
stp-ip: Not sure I like the `url = "/blog"` workaround. From an implementation perspective I see why, but from a user perspective it feels more complex than it should.
Another option would be to go with this options for now and in a separate release/step work on the defaulting.

Basically a section index, would automatically be a list fragment. Leaf bundle content would automatically be a content fragment.

layouts/partials/fragments/list.html
```diff
@@ -0,0 +1,108 @@
...
+    <div class="container py-5">
+      <l class="row">
+        <div class="col-12
+          {{- if or (eq $bg "secondary") (eq $bg "primary") -}}
+            {{- printf " text-%s" "dark" -}}
+          {{- else -}}
+            {{- printf " text-muted text-%s" "secondary" -}}
+          {{- end -}}
+        ">
+          {{- range first (.Params.article_count | default 10) $articles }}
```
stp-ip: Could be shortened to just ".Params.count" I assume.

layouts/partials/fragments/list.html
mpourismaiel: Nothing that I'm aware of. The only thing is if the list fragment is used in the root of content directory or any first level directories there wouldn't be any sections. It's just the Hugo's definition of section.

exampleSite/content/blog/_index/index.md
mpourismaiel: Changed in 9ac6dd681a5d1e8c85a0610be1fe0fe2465dad88 to `/blog/latest`

layouts/partials/fragments/list.html
mpourismaiel: Changed to `count` and also added default values in the example in b7caa1581b3a1ca548502229eef3e3efc828223e

exampleSite/content/blog/_index/index.md
stp-ip: Sorry probably could have stated my point better.
I dislike the usage of `url =` as this is something the user have to change, while knowing how Hugo would interpret it. The location was actually correct in my view.
We have `/blog/article/` so the list should be located at `/blog/`. The users just shouldn't need to know that.

stp-ip:
Adding it here as Github makes it hard to find comments, that were "fixed" via code:

> I dislike the usage of `url =` as this is something the user has to change, while knowing how Hugo would interpret it. The location `/blog/` was actually correct in my view.
> We have /blog/article/ so the list should be located at /blog/. The users just shouldn't need to know that.

mpourismaiel:
By `/blog/` do you mean `exampleSite/content/blog/` or `example.com/blog/`? If you mean the first one (directory path) If I do so then all the other directories in that directory would be considered resources for the `example.com/blog/` unless we remove all directories in that path and use files. The only way hugo allows for index page in a section is by using list layout which we cannot support due to us wanting to support fragments and list layout has a weird way of resolving files.

stp-ip:
example.com/blog/ was intended. For now it seems the workaround with `url=` is the best approach. I just wanted to mention, that it's not the best user UI. Sorry for the confusion.

stp-ip:
LTGM and iterate in follow on PRs.

PR #224, opened 2018-07-10
Add support for portfolio

### LINKED ISSUE #5: Create portfolio layout
feature

**What happened**:
No portfolio layout available.

Portfolio layout available.

Look for a portfolio layout.

Notes:
Portfolio items should have the following features:
* primary image (for things such as album cover, book cover, company)
* overlay image (ability to overlay primary image with an image)
* title
* subtitle
* summary
* linkable

Questions:
Should we also support them opening in modals instead of having a link? Something along the lines of: If link is given, click will bring you to the link. If no link is given the content will be shown in a modal with title, subtitle and image as header or similar.
Should title, subtitle and summary be defaulting to the one from linked pages? Easier to use for users, but harder to implement and more magic/complexity.

mpourismaiel:
Adds portfolio support. Modal support was added since portfolio requires it. Modal is globally available through `window.syna.showModal` so users can use it easily for their custom dialogs.

fixes #5 

- Modal is now supported and it's code lies in `static-main/js/modal.js` and it uses our `jq-helper` and is imported in `static-main/js/index.js` and there's a function in the new object `window.syna` to allow for users to open dialogs easily.
- There is a new JS chunk named portfolio which contains a small code to open the dialog containing a portfolio item's details. This chunk is loaded from the portfolio fragment.

**Release note**:
```release-note
- Add new fragment **Portfolio**
- Add support for modals
```


stp-ip:
* move the overlay etc. to the bottom 1/3
* remove button and make whole item clickable on link, for no link use modal
* if no title or subtitle -> no overlay
* add overlay image option aka ability to add an image/logo centered
* try around with various subtitle lengths - I think there is a bug for short subtitles that they flow into the line of the title

stp-ip:
Also first row of example data could be using the overlay aka having a title and subtitle.
The second row of example data could be using no title and subtitle so that the overlay isn't visible.
One of the second row items could then use markdown to emulate a title within the modal I assume.

exampleSite/content/_index/portfolio/item1.md
```diff
@@ -0,0 +1,9 @@
++++
+weight = 10
+title = "First Title"
+subtitle = "Lorem ipsum dolor sit amet, consectetur adipiscing"
+image = "cat-1.jpeg"
+link = "http://example.com"
```
stp-ip: Link to the same page `#` to prevent people leaving the demo.

stp-ip:
Things left for this to merge as PoC:
* use whole image not just title and subtitle to open modal
* assume title and subtitle are set (no overlay can be done in another iteration)
* mouse should indicate clickability for modal
* opacity to 0.6? (happy to discuss)
* use links to same page

mpourismaiel:
By links to same page, do you mean url and `#` hrefs?

stp-ip:
Just use "#" for links as we do for other urls.

PR #234, opened 2018-07-13
Remove useless bootstrap overrides and clean up variables

mpourismaiel:
Remove everything that seemed useless from `bootstrap-override` directory and the `_variables.scss` file. The resulting file seems to have the same effects as before and it's much smaller and easier to configure. 

**Release note**:
```release-note
Make the theme lighter
Better bootstrap override config
```


stp-ip:
Triggered a redeploy for netlify. Was a connection issue.

stp-ip:
Removing code \o/

PR #242, opened 2018-07-15
Fix naming consistency for logo fragment

### LINKED ISSUE #213: Get more consistent with naming
feature

**What happened**:
For buttons, logos etc. we are not yet consistent with naming.
Sometimes we use `text`, `name` or `title`.

At the very least we should just use `text` for the url or image alt texts.
If a title is actually a title we can use `title`.

Logo fragment.
Any other instances of this?

**Environment**:
- Syna Theme version: master

lauragift21:
It changes all "name" instance to "text" to enable consistent naming for all fragments.

fixes #213 

**Release note**:

```release-note
Fix naming consistency for logo fragment
```

### DIFF
--- a/layouts/404.html
+++ b/layouts/404.html
@@ -1,36 +1,36 @@
 <html lang="{{ .Site.LanguageCode | default (printf "en-us") }}">
 
-  {{- partial "head.html" . -}}
+  {{ partial "head.html" . }}
 
   <body class="bg-white">
-    {{- $global := .Site.GetPage "page" "_global" -}}
-    {{- if $global -}}
-      {{- range ($global.Resources.ByType "page") -}}
-        {{- if and (not .Params.disabled) (isset .Params "fragment") (eq .Params.fragment "nav") -}}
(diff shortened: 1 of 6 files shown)


layouts/404.html
```diff
@@ -16,16 +16,16 @@
...
-              {{- i18n "404.subtitle" -}}
+              {{ i18n "404.subtitle" }}
             </h3>
             <h5 class="my-4">
-              {{- i18n "404.direction" -}}
+              {{ i18n "404.direction" }}
             </h5>
             <a class="btn btn-primary btn-lg my-4" href="/">
-              {{- i18n "404.button" -}}
+              {{ i18n "404.button" }}
```
stp-ip: Why the change to not remove whitespace here?

stp-ip:
@mpourismaiel we should probably do the contact refactoring in a follow on PR. Can you open an issue or PR to track that.

layouts/404.html
lauragift21: Oh, I removed this here because I created a branch off my previous PR and had the changes reflect here I thought it will cause some kinda merge conflict. I'll add it back.

lauragift21:
Thanks @mpourismaiel for pointing that one out. Since @stp-ip suggested it should be moved to a new issue.

PR #249, opened 2018-07-18
Fix button alignment in item fragment when there is an icon

### LINKED ISSUE #202: [item] Button not centered when align = center
bug

**What happened**:
Buttons are left aligned, when alignment for item fragment is set to "center" and an icon is used.

Center buttons.

**Environment**:
- Syna Theme version: v0.11.0

mpourismaiel:
Fix button alignment in item fragment when there is an icon

image

fixes #202 

Item fragment is too complicated. Let's consider simplifying it.

**Release note**:
```release-note
bug: Item fragment didn't align buttons when there was an icon present
```

### DIFF
--- a/layouts/partials/fragments/item.html
+++ b/layouts/partials/fragments/item.html
@@ -169,13 +169,13 @@
                   {{- printf " text-muted text-%s" "secondary" -}}
                 {{- end -}}
               ">
-                {{- .self.Content  | markdownify -}}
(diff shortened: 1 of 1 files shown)


stp-ip:
Let's move the simplification discussion into a separate issue. Happy to discuss it in more detail. It definitely should duplicate as much code as it does.

PR #251, opened 2018-07-18
Fix 404 page logo alt text

### LINKED ISSUE #248: [404] improve handling of no logo
bug

**What happened**:
Currently if no logo is defined/available it will show image not found icons.

Not sure what to expect, but maybe we could fallback to text or so.

mpourismaiel:
Using alt text on the image seems like a good idea for this issue. It's more accessible as well.

image

fixes #248

### DIFF
--- a/layouts/404.html
+++ b/layouts/404.html
@@ -13,21 +13,21 @@
     {{- end -}}
 
     <section>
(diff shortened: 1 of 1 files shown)


mpourismaiel:
The borders around the screenshot is the magic of ImageMagick.

PR #254, opened 2018-07-20
Change behavior of fragment and children selection

### LINKED ISSUE #179: Overwriting global fragments doesn't work with subpath.
bug

**What happened**:
Using a subpath the global fragment can't be overwritten.

`/content/index/nav/index.md` would not overwrite `/content/global/nav.md`

Usually a global fragment can be overwritten with a fragment with the same file name.

I assume we are checking against *.md and not just *.

mpourismaiel:
Fixes two issues:
- Fragments with children such as item were allowed to be placed in the root directory (`%page/%fragment.md` with children in `%page/%fragment/%children.md`) but if the fragment was placed inside that directory it would be problematic. This is now fixed.
- Global fragments were allowed to render if the local fragment was placed inside a directory even if that directory had the same name with the local fragment. This is fixed as well.

fixes #179 

**Release note**:
```release-note
- BREAKING: Global fragments will not be rendered if there is a local fragment inside a directory and the directory has the name of the global fragment.
- Fragment content files can now be placed inside a directory even if they have children.
```


stp-ip:
Awesome. Now we are consistent with Hugo and the other fragments \o/

PR #255, opened 2018-07-20
Add registry for js files in fragments

mpourismaiel:
Loading js files in fragments was not a good idea since adding several of a single fragment would mean the js file is required several times. This way we can require it only once.

**Release note**:
```release-note
Optimize how js files are handled
```


stp-ip:
Interesting approach. Seems like a good solution. Would love another review from @lauragift21 on this.

Needs a rebase for the conflict.

layouts/partials/fragments/contact.html
```diff
@@ -53,7 +57,7 @@ <h2>{{- . | markdownify -}}</h2>
                       type="text"
                       data-validation
                       data-validation-error-msg="{{ .error | default (i18n "contact.defaultNameError") }}"
-                      placeholder="{{ with .button_text }}{{ . | markdownify }}{{ end }}"
```
lauragift21: Did we decide with "text" again? thought we had it changed.

stp-ip: No afaik we decided on .button_text. Not sure what happened here. Maybe just some rebase hiccup. @mpourismaiel

mpourismaiel: Because this file contained `text` and I thought that was correct one. I'll change it ASAP.
https://github.com/mpourismaiel/syna/blob/d9b291ee3028849a7bcc7fd3e3d2ed78d688a186/exampleSite/content/_index/contact.md

mpourismaiel: Actually I think `text` is more accurate in this case since `buttent_text` is not descriptive for the input's placeholder.

stp-ip: Ahh you are correct. The placeholder is actually .text.

.button_text was just the button. Totally missed that.

PR #259, opened 2018-07-22
Refactor 404 layout

mpourismaiel:
Make 404 page support fragments by creating a 404 fragment (same as the previous layout) and adding that fragment to the global fragments. 404 fragment will not be loaded anywhere except the 404 page if it's added in the global directory.

**Release note**:
```release-note
404 page is now customizable using fragments
```

### DIFF
--- a/exampleSite/content/_global/404.md
+++ b/exampleSite/content/_global/404.md
@@ -1,0 +1,11 @@
++++
+date = "2018-07-22"
+fragment = "404"
+weight = 150
+
+#title = "" # default i18n "404.title"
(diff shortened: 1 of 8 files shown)


stp-ip:
Not sure I like the special casing of `fragment = 404` only rendering on the 404 page. That feels imperfect. But happy to merge this now and reiterate with more production sites using it. Aka getting a feel on the real world implications.

The rest seems good. Needs documentation so in a follow on PR.
* [ ] document 404 fragment special case
* [ ] document global only fragment support

PR #260, opened 2018-07-22
wip: Add hugo resource pipeline to minify resources

mpourismaiel:
Use hugo resource pipeline for scripts and styles. Moved scripts and styles from `static` to `assets` directory since pipeline requires it.

I'm planning on concatenating scripts as well but right now there seems to be an issue either with the code I stashed or with Hugo.

**Release note**:
```release-note
Minified scripts and styles
```


mpourismaiel:
Fonts are not being placed in the right directory right now. And Hugo assets directory has a weird behavior. Needs to wait until further improvement.

stp-ip:
Looks good so far. Will wait for the final iteration.

mpourismaiel:
New versions of the Hugo on aur/hugo-git seem too buggy. There are some problems I've encountered. For example current iteration of pipelines don't take urls into account. So for example a font should be imported manually whereas webpack would extract the font into the given directory. I think it's a long way until pipelines would be usable and a better alternative than webpack.

PR #271, opened 2018-08-06
Update to v0.12 with latest changelog

stp-ip:

**Release note**:

```release-note

```


CHANGELOG.md
```diff
@@ -1,15 +1,51 @@
...
 ## Documentation for v0.12.0
-[Documentation](/tree/v0.11.0/docs)
+[Documentation](/tree/v0.12.0/docs)
+[Getting started](/tree/v0.12.0/docs#using-starter)
 
 ## Changes since v0.11.0
+  - BREAKING: `content-single` and `content-split` merged into `content` fragment
+  - BREAKING: Moving to `_index` and `_global` as special directories and headless bundles
+  - BREAKING: Subpath handling made consistent with Hugo
+  - New: `header` fragment for easier section bundling and linking
```
mpourismaiel: `header` -> `section-header`

Should I rename it to `header`?

CHANGELOG.md
```diff
@@ -1,15 +1,51 @@
...
+  - Bundle JS files and register them within each fragment
+  - `404`: Refactor 404 to be fragment based
 
 ## Fixes since v0.11.0
+  - Improve naming consistency
+  - Cleanup bootstrap files
+  - Add attribution for inspiration
+  - `table`: Align table cells using `align` variable
+  - `items`: Remove icon, if not set
+  - `item/table`: Fix icon
```
mpourismaiel: URL is fixed as well for item and table but not sure if it's necessary to point out.

mpourismaiel:
Did they change review comment? Anyways there's `package.json`'s version as well. I don't think it can be removed so keeping that in sync may be a good idea.

CHANGELOG.md
stp-ip: Mhh might make sense to just go with header?

CHANGELOG.md
stp-ip: I'll add it.

stp-ip:
Ah yeah. We definitely need release documentation to know, what needs to be changed. Less "knowing" the specifics and more of a checklist type thing.

CHANGELOG.md
stp-ip: Added a migration commit. Any arguments against going with the simpler one aka `header`?

mpourismaiel: I think we had one but don't remember the outcome. Header sounds good.

PR #288, opened 2018-08-12
Add dev examples for images not loading and unconfiged ones

stp-ip:

**Release note**:

```release-note

```

### DIFF
--- a/exampleSite/content/dev/missing-images/footer.md
+++ b/exampleSite/content/dev/missing-images/footer.md
@@ -1,0 +1,17 @@
++++
+fragment = "footer"
+date = "2016-09-07"
+weight = 1200
+
+[logo]
+  title = "Logo Title"
+  image = "404.png"
+  text = "Logo Subtext"
+  url = "#"
++++
+
+#### Description Title
+
+Project description such as:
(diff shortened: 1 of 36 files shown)


stp-ip:
The case for not loading images aka bad connection: https://5b703784c96592748e389ebe--syna-demo.netlify.com/dev/missing-images/

The case for unconfigured images: https://5b703784c96592748e389ebe--syna-demo.netlify.com/dev/unconfigured-images/

With this demo we can create a few issues and fix both cases to make them look good.

mpourismaiel:
Fantastic. For missing images, should we display an error message or render it as is?

stp-ip:
I would say for non configured images, render as is. There might be a specific reason for that and as long as it doesn't break the layout, that should be ok.

mpourismaiel:
@stp-ip What about missing images? For non configured we mostly don't render the `img` tag altogether (logos and portfolio fragment require images so doesn't make sense to remove them, just make them better looking).

stp-ip:
Missing images I think it maybe should fallback to sort of a blank placeholder aka keep the layout as if an image would be present. Such as an empty circle for member etc.

PR #290, opened 2018-08-12
Add alignment variables for header fragment

### LINKED ISSUE #283: Add alignment to header fragment
feature

**What happened**:
Header fragment is only displayed in center.

To be able to align the header.

Check out the header fragment.

**Environment**:
- Syna Theme version: v0.12

mpourismaiel:
Header fragment now supports alignment

fixes #283 

**Release note**:
```release-note
- header: Added support for alignment of title and subtitle
```

### DIFF
--- a/exampleSite/content/_index/header.md
+++ b/exampleSite/content/_index/header.md
@@ -1,0 +1,11 @@
++++
+date = "2018-07-13"
+fragment = "header"
+weight = 500
+
+background = "white"
+title = "Section Header Fragment"
+subtitle = "Even linking is possible. This fragment can be used for related sections so linking is easier."
+#align = "left" # Default center, available values: left, center, right, with bootstrap breakpoint variables (default: xs, sm, md, lg, xl)
+#subtitle_align = "left" # Default to align value, available values: left, center, right, with bootstrap breakpoint variables (default: xs, sm, md, lg, xl)
++++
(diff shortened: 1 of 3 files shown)


exampleSite/content/_index/header.md
```diff
@@ -0,0 +1,11 @@
++++
+date = "2018-07-13"
+fragment = "header"
+weight = 500
+
+background = "white"
+title = "Section Header Fragment"
+subtitle = "Even linking is possible. This fragment can be used for related sections so linking is easier."
+#align = "left" # Default center, available values: left, center, right, with bootstrap breakpoint variables (default: xs, sm, md, lg, xl)
```
stp-ip: "exposing" the breakpoint variables seems imperfect. Yes it works, but is that the configuration interface we want to expose? Especially through "align"?

mpourismaiel: Since we're doing `text-%s` it's possible to use `text-md-center` for example since Bootstrap allows it. I can remove the documentation but seemed better to have it.

stp-ip: Mhh I would leave it out for now. Yes we can do that, but I would discuss that separately. Especially as we don't do it with any other alignment yet. It would be a break of consistency. Let's open a separate issue for that and discuss.

mpourismaiel:
I removed the comment regarding the breakpoints.

PR #319, opened 2018-08-20
Add bootstrap unresponsive grid to head of every page

### LINKED ISSUE #264: Minimal inline CSS
feature

**What happened**:
Currently the page loads CSS asynchronously. This might lead to unstyled html output, which looks suboptimal before the CSS is loaded.

Minimal inline CSS styling to emulate some form of structure before CSS is loaded.

mpourismaiel:
Extracted bootstrap grid and removed any responsive code and added the result to exampleSite. The style is loaded inside the head.html partial file in order to have a readable website in case the main css file fails.

fixes #264 

I'd like to wait and see what Hugo would have to offer for this one. I'm not sure the behavior of readFile is buggy, and I'm not sure expecting a resources.Content would be a good thing. But maybe we can revisit the issue at the end of the milestone?


stp-ip:
Looks good. Minimal css to make it look ok-ish, if main.css fails. Also doesn't add that much of an overhead.
A few issues I see.
We do that with a file in "user space". This is due to readFile limitations, but we should try to work around that.
Not sure on what you want to wait, but I'm happy to wait for the end of the milestone on this.

mpourismaiel:
I'd like to wait and see if `readFile` would receive any improvements. Although it's possible to minify the css and append it to `head.html` but that would be way too messy.

mpourismaiel:
Is there anything else we can do with this?

stp-ip:
Not really. One option would be to open an issue/feature request to allow readFile to have a theme equivalent such as readThemeFile or something. Happy to let the team work on such a feature. Would that help in the long run?

stp-ip:
Sounds good. As you have the view of what readThemeFile might look like from a syntax perspective and what the actual needs are. Happy for you to create a proposal.

mpourismaiel:
There is already an issue open, pending from v0.43 https://github.com/gohugoio/hugo/issues/4842

stp-ip:
So it seems we could already make it work, but sort of hardcoded. Would that work as a tmp solution, with an issue tracking the hugo issue and fixing it, when we can?

mpourismaiel:
The tracking issue is available on gohugo, we can use this solution for now and create another issue to refactor and start it whenever hugo complies.

PR #342, opened 2018-09-04
Update CHANGELOG for v0.13.0

stp-ip:
Update Changelog including all changes for v0.13.0 release.


stp-ip:
Merging now. Didn't find any breaking changes yet. Will do a follow-on PR, if I do. Seems like we are actually getting to less and less breakage \o/

PR #346, opened 2018-09-05
Fix image alignment in member fragment single mode

mpourismaiel:
Single member in member fragment didn't position the image correctly, also the example was removing some data.


stp-ip:
*cough* #344 *cough* :smile:

PR #359, opened 2018-09-11
Add support for asset.icon where possible and add images to items fragment

### LINKED ISSUE #325: [logos, items] There is no fragment to display an image with text
feature

**What happened**:
Neither items nor logos fragment allow for image with text.

To be able to implement something like the following:

http://pollylabs.org/partners.html

mpourismaiel:
- .Params.asset.icon is now possible in item and items fragment
- Deprecate logos fragment and every reference of it
- Add support for image in items fragment's columns
- Add items fragment to dev section (unconfigured-images, missing-images, resource-fallthrough)
- Add examples of items fragment with image and content and with only image

fixes #325 

This is quite a big change. I reviewed it myself so to see whether I've missed anything. The changes to fragments/ examples should be reflected in code-* files as well.

- In navbar there are three assets (logo, search, repo) and renaming to asset was not possible.
- In members fragment there are two assets (asset for image and icons for social links generally) and renaming to asset was not possible.
- In table and item fragment, inside the table there is an icon but I'm not sure how to make that a toml table :D

**Release note**:
```release-note
- Breaking: item: Icons should be placed under `asset` table
- Breaking: items: Icon for each column should be placed under `asset` table
- Breaking: logos: Deprecated `logos` fragment in favor of `items` fragment
```


mpourismaiel:
I feel icon under asset makes the configuration easier and more understandable. We only need to document how to define an asset and which fragment comes with what variables. That would be it. I think that's easier.

stp-ip:
Overall good.

A few things that feel a bit weird in items is that the url is set outside the asset and only links the title.
Maybe it should be set in the [asset] and link both the image/icon and the title?

Also the logo example doesn't showcase linking an image, which was a feature of logos.

mpourismaiel:
Did the following, hope I didn't miss anything.

- Fix url not located inside the asset
- Add url to title and to icon
- Remove title from html when it's not provided
- Remove content from html when it's not provided
- Add more examples

exampleSite/config.toml
```diff
@@ -8,7 +8,6 @@ themesDir = "../.."
 enableGitInfo = true
 version = "0.13.0"
 disableKinds = ["RSS", "taxonomy", "taxonomyTerm"]
-paginate = 2
```
stp-ip: What's the reason for removing paginate here?

mpourismaiel: No reason actually but it was a priority fix so removing it asap seemed like a good idea. Should I move it to a patch PR?

stp-ip: Was just a bit out of context. Let's try to do that in separate PRs to keep the context of PRs. We can leave it in here this time.

PR #370, opened 2018-09-16
Add alignment support to header helper

mpourismaiel:
Header helper doesn't support alignment. This PR adds the previous helper fragment features to the header helper and supports alignment.

**Release note**:
```release-note
- global: All fragments now support alignment of the title and subtitle
- BREAKING: header: `align` variable is changed to `header_align`
```


stp-ip:
A few things:
* Release note is about "header_align", which isn't used in the code.
* Do we really want to split it up and make both subtitle and title configurable separately? Feels like too much flexibility.
* Helper should also be used within the header fragment.

PR #374, opened 2018-09-18
WIP: Chore/colors/create dev

mpourismaiel:

**Release note**:

```release-note

```

### DIFF
--- a/layouts/partials/helpers/text-color.html
+++ b/layouts/partials/helpers/text-color.html
@@ -1,12 +1,13 @@
 {{- $bg := .self.Scratch.Get "bg" }}
 {{- $text_muted := cond (.muted | default false) "text-muted " "" -}}
-
+{{- $text_dark := cond (.dark | default false) (printf "text-%s" (.dark | default "dark")) "" -}}
+{{- $text_light := cond (.light | default false) (printf "text-%s" (.light | default "light")) "" -}}
 {{- if .overwrite -}}
   {{- printf " text-%s" .overwrite -}}
 {{- else -}}
   {{- if or (eq $bg "white") (eq $bg "light") (eq $bg "secondary") (eq $bg "primary") -}}
-    {{- printf " %stext-%s" $text_muted (.dark | default "dark") -}}
+    {{- printf " %s%s" $text_muted $text_dark -}}
   {{- else -}}
-    {{- printf " %stext-%s" $text_muted (.light | default "light") -}}
+    {{- printf " %s%s" $text_muted $text_light -}}
   {{- end -}}
 {{- end -}}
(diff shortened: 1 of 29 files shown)


layouts/partials/helpers/text-color.html
```diff
@@ -0,0 +1,13 @@
+{{- $bg := .self.Scratch.Get "bg" }}
+{{- $text_muted := cond (.muted | default false) "text-muted " "" -}}
```
stp-ip: Muted does mostly make sense on light backgrounds. Assuming we have to codify the different variations anyway, I think it's easier to remove muted. What do you think?

layouts/partials/helpers/text-color.html
```diff
@@ -0,0 +1,13 @@
...
+{{- $text_dark := cond (isset . "dark") (cond (eq .dark "") "" (printf "text-%s" .dark)) "text-dark" -}}
+{{- $text_light := cond (isset . "light") (cond (eq .light "") "" (printf "text-%s" .light)) "text-light" -}}
+{{- if .overwrite -}}
+  {{- printf " text-%s" .overwrite -}}
+{{- else -}}
+  {{- if or (eq $bg "white") (eq $bg "light") (eq $bg "secondary") (eq $bg "primary") -}}
+    {{- printf " %s%s" $text_dark -}}
```
stp-ip: `{{- printf " %s" $text_dark -}}`

layouts/partials/helpers/text-color.html
```diff
@@ -0,0 +1,13 @@
+{{- $bg := .self.Scratch.Get "bg" }}
+{{- $text_muted := cond (.muted | default false) "text-muted " "" -}}
+{{- $text_dark := cond (isset . "dark") (cond (eq .dark "") "" (printf "text-%s" .dark)) "text-dark" -}}
+{{- $text_light := cond (isset . "light") (cond (eq .light "") "" (printf "text-%s" .light)) "text-light" -}}
```
stp-ip: What's the idea behind setting .dark and .light? Shouldn't that be something that the helper figures out from the background?
With a manual overwrite say:
```
{{- partial "helpers/text-color.html" (dict "self" $self "overwrite" "primary") -}}
```

Where all text-colors would be set to primary with the exception of a primary background.
With a primary background it would be an dark for example. Does that make sense?

assets/styles/_syna.scss
```diff
@@ -63,3 +63,13 @@ article .content {
...
   }
 }
+
+pre {
+  padding: .75rem;
+}
+
+.bg-dark {
+  pre {
+    background-color: white;
```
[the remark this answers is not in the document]
mpourismaiel: Weirdly enough bootstrap with it's hundreds of variables doesn't provide any variable or styling for code blocks. And the only play the code blocks are not readable are in a dark background. Since we don't control the rendering of code blocks (since user enters it in markdown) the only place to customize it is in css.

mpourismaiel:
I'm not sure how to customize the link color in primary background and `:hover` color in dark backgrounds. The only way would be css but what color should I set? We can use inline code color but that doesn't have any `:hover`.

mpourismaiel:
Is there anything we can do on this PR?

PR #379, opened 2018-09-25
Add new yarn test:colors and the generated files

mpourismaiel:
- Added new _tests directory, containing a script to copy fragments to `dev/colors` section
- Create every fragment with every color inside `dev/colors`

**Note:** Files of note:
- _tests/colors-generator.js
- package.json

Other files are generated from fragments section. Only their weight and their background colors are different.


stp-ip:
Great. But it merges all fragments and it's colors into one directory with weird ordering.
In my view it should keep the per fragment subdirectory and then in each fragment showcase the colors. In the same order.

What do you think?

mpourismaiel:
Sorry but I don't understand what you mean. Would you please elaborate?

stp-ip:
Instead of having all color examples unordered in `/dev/colors/` the idea is to have it similar to /fragments/ such as `/dev/colors/fragments/item/`. So on this item page it would be easy to showcase the different colors ordered specifically.

mpourismaiel:
Oh so you mean to add each fragment/color into it's own page? That would be fantastic for showcasing but what about testing and checking if a new color is breaking something?

stp-ip:
Yes we can easily more that idea for showcasing afterwards.
testing is a bit harder, but also more organized. If one changes the item fragment one doesn't have to look through all other fragments. Only for color changes one needs to check every item.
So I would go with let's do it with subdirectories in dev. Test it for a release and then see how to move it to a showcase section. Does that make sense?

stp-ip:
General sidenote: The content left alignment with not a lot of text looks a bit weird.

Some things:
* [ ] buttons: need a subtitle
* [ ] buttons: we should test at least all non outline colors
* [ ] content: centered one needs subtitle
* [ ] editor: broken (not showing up)
* [ ] embed: would just use the iframe for simplicity
* [ ] hero: remove header image
* [ ] items: would just showcase icon and image both with title, subtitle and text
* [ ] list: subtitle?
* [ ] nav: add search and breadcrumb
* [ ] react: broken (not showing up)
* [ ] search: subtitle?

For all fragments, that have various options it might make sense to group them by type. Say for content left we generate all colors. Then for content center and then for content right.

That would make differences for each between different colors easier to catch.

exampleSite/content/fragments/search/search.md
```diff
@@ -4,7 +4,7 @@ weight = 110
 title = "Search"
 background = "secondary"
 
-#title = ""
-#subtitle = ""
+title = "Search Fragment"
+subtitle = "Searches through all content fragments"
```
stp-ip: Both title and subtitle aren't used on the rendered page aka probably not used within the layout yet.

mpourismaiel:
@stp-ip I don't understand the issue with search fragment. I can see both title and subtitle:

image

I updated the dev/colors section with new faq fragment, updated examples. Didn't update item fragment since it's being copied from the fragments section so it would be hard to change. Can we keep that as is?

stp-ip:
Yeah resolved. Didn't see the regenerated color fragments yet.

stp-ip:
Updated the items still open.

Items feels too crowded and not necessary. Couldn't we selectively only copy over a specific fragment. Say copy over "items" and "logos", but not the others?

Same for embed to just copy over the iframe to simplify testing etc.

Hero is harder to test with the background image. We could cheat a bit and have the background images colocated with the page or fragment and just not copy over the background for the color section therefore it would 404, but that feels like a dirty hack.

IDs would need to be fixed, when we use it as actual color showcase. If we agree, that we can merge before. We should create an issue to track it so.

mpourismaiel:
I don't understand what you mean by ID.

For items fragment and others like it, I can add an exception object in the script but that would need maintaining. Would that be fine?

stp-ip:
Whitelists are easier to maintain, but if the code would get much more complex, let's go with an exception object instead.

collapse and uncollapse use the `id=` of the element. These are the same for all colors and therefore break the functionality.

mpourismaiel:
Mehrnoosh had it fixed by adding fragment directory name prepended to the ID. She removed it in favor of easier linking. Should I add it back?

stp-ip:
Usually one wouldn't have the same filename in the same directory so easier linking makes more sense, but in this case it breaks functionality. It's a special case so happy to open an issue and fix it, when we move colors to be showcased. Makes sense?

mpourismaiel:
Fantastic.

mpourismaiel:
The same thing happens for nav search as well. Should we create an issue for that as well?

stp-ip:
Yeah mentioned it already in the list of items. Will mark both of them checked, when we have separate issues. :+1:

mpourismaiel:
Fixed:
- buttons now have outline
- buttons now have all supported colors
- items now only shows items and logos, with title and subtitle and content
- embed now only shows iframe

Not sure how to remove background image from hero without too much customization in the script.

Also split the script a little for better readability. Not much readable however. Sorry :D

stp-ip:
Hehe wasn't sure, if it's just me not knowing javascript that good :)

Ok. Let's merge and create the issues as discussed to move the last items to their own issues. We can leave the hero image for now, but would be great, when you create an issue for it. So I might take a look at a hacky non js solution.

PR #380, opened 2018-09-25
Fix display of images in list fragment and add examples

mpourismaiel:
Images of content in pages were not being shown in the list fragment.

**Release note**:
```release-note
- list: Fix images not showing up
```


stp-ip:
That will always show images in list (non tiled mode) with no way to disable that. It's a good default, but what do you think about an option to disable images?

mpourismaiel:
Sounds good. Will add it right away.

PR #382, opened 2018-09-25
Add breadcrumbs support to nav fragment

mpourismaiel:
This PR adds support for breadcrumbs in the navbar and enables it by default in exampleSite.

**Release note**:
```release-note
- nav: Two new variables added enabling support for breadcrumbs:
	- breadcrumb: default false
	- breadcrumb_level: default 1
```


mpourismaiel:
I think it would be a good idea to add an option to customize the background color of the breadcrumb, or even separate the breadcrumbs from navbar. Not really sure though. What do you think @stp-ip ?

stp-ip:
Happy for the background to be overwritable so it might move more into the background.

For the current design. I would use a smaller font size and a margins. It's too big. Probably enough to have it half the height of the navbar. What do you think?

Also the breadcrumb bar currently overlays the search ;)

mpourismaiel:
Haha, will fix these tomorrow.

exampleSite/content/_global/nav.md
```diff
@@ -5,8 +5,11 @@ date = "2018-05-17"
 weight = 0
 background = "dark"
 search = true
-breadcrumb = true # Default value is false
-#breadcrumb_level = 0 # Default is 1
+
+[breadcrumb]
+  display = true # Default value is false
+  level = 0 # Default is 1
```
stp-ip: What does `0` do?

layouts/partials/fragments/nav.html
```diff
@@ -98,17 +98,21 @@
     </div>
   {{- end }}
 </nav>
-{{- if and .Params.breadcrumb (lt (.Params.breadcrumb_level | default 1) (sub (len ($sections | default slice)) (cond (eq $.root.Kind "home") 1 0))) }}
-  <nav aria-label="breadcrumb" class="overlay navbar navbar-expand-lg
-    {{- printf " bg-%s" $bg -}}
+{{- $breadcrumb := .Params.breadcrumb -}}
+{{- $display_breadcrumb := and $breadcrumb $breadcrumb.display -}}
```
stp-ip: Why the specific `and` usage here?

layouts/partials/fragments/nav.html
```diff
@@ -98,17 +98,21 @@
...
+{{- $level := $breadcrumb.level -}}
+{{- $nav_bg := $breadcrumb.background | default "light" -}}
```
stp-ip: Reminds me of our background non helper days ^^

exampleSite/content/_global/nav.md
mpourismaiel: I commented it out. It's the level from which we show the crumbs. Since breadcrumbs are displayed in order to show parents, I provided this variable to customize from which parent we should display the crumbs. For example home is always at the level 0, every page in root section is level 1. If those pages are sections containing other pages, those pages are level 2 and so on.

layouts/partials/fragments/nav.html
mpourismaiel: I rewrote this entire part of code. It was entirely wrong :D

layouts/partials/fragments/nav.html
mpourismaiel: Yeah, since we removed the background helper this was necessary :D

exampleSite/content/_global/nav.md
stp-ip: Ah I was thinking at first thought it's the depth level.

stp-ip: Probably best to add a comment here.

layouts/partials/fragments/nav.html
stp-ip: Did I miss something. What specifically did we remove?

stp-ip:
2 things still:
* On the home page we shouldn't show anything.
* With hero it looks weird. Maybe the breadcrumb span or so should just overlay instead of having a whole navbar.

layouts/partials/fragments/nav.html
mpourismaiel: The `background` helper. We removed it, moved the snippet to `fragments` helper, added `bg` to `Scratch` and used the value inside `text-color` helper.

mpourismaiel:
First bullet is fixed now. For the second one, I think the better fix would be a breadcrumb fragment instead of breadcrumb in navbar. Overlaying it would then hide part of hero behind it.

stp-ip:
Mhhh depends on the hero. On wider displays it wouldn't be problematic, but on smaller ones it would. Then it probably should be the same background as the nav. At least the whole navbar. Not the span. Or else it looks weird in my view. What do you think?

stp-ip:
screenshot from 2018-09-28 22-20-30

stp-ip:
But maybe that's just me thinking it feels weird. Maybe merge and reiterate?

mpourismaiel:
I still think a separate breadcrumb fragment would be better. We can place it under the hero or any other fragment in that case. With breadcrumbs inside navbar we would need to check every page to see if it renders fine. For now however, we can merge and see if it would be necessary to reiterate.

stp-ip:
Yeah let's merge for now and reiterate. Another fragment seems too much complexity for something like that.

layouts/partials/fragments/nav.html
stp-ip: Wouldn't that mean that we could get the bg from scratch instead of defaulting here again?

PR #383, opened 2018-09-29
Fix fragments lookup order

mpourismaiel:
The lookup strategy was top to bottom, making the fragments in higher levels more important than fragments closer to the page. So fragment override didn't work as expected. Also directory name was not checked in cases of fragments located in directories. This patch fixes that as well.

@stp-ip This patch most likely will break some pages. If there are any pages that have a fragment that has the same name or dir_name as one in a parent section's global, the fragment closer to page is rendered and will override the other fragment which is further away from the page. I broke the lookup when I was trying to add global fragments in every section. Sorry.

**Release note**:
```release-note
- Breaking: Fragment lookup order was broken in the last version. The lookup was done from root to page where it should have been page to root (page > parent > ... > root). This patch will fix the order but may break the pages where a fragment was meant to override another fragment with the same name or directory name.
```


stp-ip:
Let's add a test section for this using the copyright once on a section page and another time on a normal page.
Does that make sense?

mpourismaiel:
Will do.

stp-ip:
Should we also add a test for section level overwrites?

PR #394, opened 2018-10-01
Enable slots (fragment in fragment) 

mpourismaiel:
This PR provides means to register slots within any fragment (currently list and content have slots) and assign other fragment to those slots, allowing for much more complex and customizable designs while keeping the simplicity of Syna.

New approach for Fragment-in-Fragment. This way we register slots in any fragment and assign a fragment to that slot by addressing it in frontmatter. For example `slot = content/sidebar` along with `fragment = list` would add a list fragment to sidebar slot of a `content.md` fragment.

This would be a really good approach. As we're close to v0.15 I think considering it sooner would be a good idea.

**Release note**:

```release-note

```


exampleSite/content/dev/blog/_global/sidebar-list.md
```diff
@@ -0,0 +1,7 @@
++++
```
stp-ip: From a design perspective. The fragment in fragment has too much whitespace around it also it's currently not usable. Images shouldn't be there. Header text is too big etc.

That gives me a bit of a reason to think it's hard to optimize.

layouts/partials/fragments/list.html
```diff
@@ -36,7 +36,9 @@
       <div class="row mx-0
         {{- partial "helpers/text-color.html" (dict "self" $self.self "light" "secondary") -}}
       ">
+        {{ partial "helpers/slot.html" (dict "root" $ "slot" "before-all")}}
```
stp-ip: `before-all` seems incorrect as it's after the header, but `before-content`.

layouts/partials/fragments/list.html
```diff
@@ -36,7 +36,9 @@
...
         {{- range first (.Params.count | default 10) (.page_scratch.Get "pages") -}}
+          {{ partial "helpers/slot.html" (dict "root" $ "slot" "before-item")}}
```
stp-ip: Why the specific `before-item`? What's the use case?

layouts/partials/fragments/list.html
```diff
@@ -149,8 +151,10 @@ <h5 class="col-12 col-md px-0">
           {{- if $self.Params.tiled }}
             </div>
           {{- end -}}
+          {{ partial "helpers/slot.html" (dict "root" $ "slot" "after-item")}}
         {{- end }}
       </div>
+      {{ partial "helpers/slot.html" (dict "root" $ "slot" "after-all")}}
```
stp-ip: `after-content`?

layouts/partials/fragments/list.html
```diff
@@ -149,8 +151,10 @@ <h5 class="col-12 col-md px-0">
           {{- if $self.Params.tiled }}
             </div>
           {{- end -}}
+          {{ partial "helpers/slot.html" (dict "root" $ "slot" "after-item")}}
```
stp-ip: Difference between `after-item` vs `after-all`

exampleSite/content/dev/blog/_global/sidebar-header.md
```diff
@@ -0,0 +1,6 @@
++++
+fragment = "index/sidebar:header"
```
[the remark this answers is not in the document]
mpourismaiel: It is the fragment name. In blog we use `index.md` files as `content` fragments.

exampleSite/content/dev/blog/_global/sidebar-list.md
mpourismaiel: The whitespace can be stripped out both by CSS or by extracting all the `.container`, `.container-fluid` and `.overlay` into a helper and call use the helper if the fragment is not in a slot. Other variables can be set to false if the fragment is registered as a slot or user can disable them manually.

layouts/partials/fragments/list.html
mpourismaiel: It can be used if the user wants to display author name, categories or any other meta data before an article.

exampleSite/content/dev/blog/_global/sidebar-list.md
stp-ip: I think then a helper makes much more sense. As we want to have the option to have a stripped render mode for fragments or such.

layouts/partials/fragments/buttons.html
```diff
@@ -1,23 +1,16 @@
...
-          <a class="btn btn-lg m-2{{ if hasPrefix .url "#" }} anchor{{ end }}
-            {{- $color := .color | default "primary" -}}
-            {{- printf " btn-%s" $color -}}
-          " href="{{ .url | relLangURL }}">
-            {{- .text -}}
-          </a>
-        {{- end }}
-      </div>
-    </div>
+{{- partial "helpers/container.html" (dict "start" true "fragment" "Buttons" "in_slot" .in_slot "bg" $bg "Name" .Name "padding" .Params.padding) -}}
```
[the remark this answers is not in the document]
mpourismaiel: That is used to render this comment: https://github.com/okkur/syna/blob/master/layouts/partials/fragments/buttons.html#L4
It would be useful for debugging the layout if someone wants to debug using source code.

mpourismaiel:
The subscribe's height is coming from the config in the embed fragment. I set the height to a bit more that what we have in the showcase page since here the width is less and causes the text to break. Should I increase it a bit more?

stp-ip:
Let's increase it a bit more and fix this minor thing in this PR.

layouts/partials/fragments/buttons.html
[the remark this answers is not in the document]
mpourismaiel: Makes sense.

PR #403, opened 2018-10-08
Activate rss feed

### LINKED ISSUE #262: RSS feed
feature

**What happened**:
No global nor per section feed is available.

Per section and one global feed available via RSS.

mpourismaiel:
Activate rss files

fixes #262 

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -7,10 +7,10 @@ theme = "syna"
 themesDir = "../.."
 enableGitInfo = true
 version = "0.14.0"
-disableKinds = ["RSS", "taxonomy", "taxonomyTerm"]
+disableKinds = ["taxonomy", "taxonomyTerm"]
 
 [outputs]
-  home = ["HTML", "JSON"]
+  home = ["HTML", "JSON", "RSS"]
 
 [frontmatter]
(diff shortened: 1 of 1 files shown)


stp-ip:
Taking a look at the actual index: https://deploy-preview-403--syna-demo.netlify.com/index.xml
I think a global index doesn't really make sense. Should updates to the page really be within the rss feed?
Maybe we need config to only have section based feeds?
Or do you think global is a good default and then we can add configuration to make the feed less spammy?

stp-ip:
Ahh ok. Mhh so hugo does have the global rss and the per section rss...
https://deploy-preview-403--syna-demo.netlify.com/dev/blog/index.xml
That one actually looks good.

mpourismaiel:
I'm not really familiar with how a feed should look like. This is done by Hugo's default RSS template.

stp-ip:
Let's say usually articles are included in the feed, but static pages and their updates would most likely not.
Let me think about this for a bit.

stp-ip:
Idea:
* add `rss = true` on the section level
* use rss template to link to the page of the content not to a single content fragment
* use rss template to show the description of the page or the first content fragment
* global rss index as a combination of the various section rss pages
* per section rss index

PR #410, opened 2018-10-16
Fix contact form

stp-ip:
Parly fixes netlify contact form submissions.

The netlify recaptcha configuration was a mixture of custom recaptcha and netlify injected recaptcha before. This is moved into the correct configuration for the netlify injected recaptcha.

Currently still a problem is that netlify seems to inject an additional recaptcha after submission aka it seems that it's ignoring the first recaptcha response, which is actually send in the correct POST request.

Also still annoying is the thank you notice being delivered out of page instead of inline.

**Release note**:

```release-note

```

### DIFF
--- a/layouts/partials/fragments/contact.html
+++ b/layouts/partials/fragments/contact.html
@@ -19,9 +19,9 @@
           class="contact"
           method="POST"
           name="{{- .Params.form_name | default (printf "contact-form-%s" .Name) }}"
-          id="{{ printf "contact-form-%s" .Name }}"
+          id="{{- .Params.form_name | default (printf "contact-form-%s" .Name) }}"
           {{- if .Params.netlify -}}
-            {{- safeHTMLAttr (print " netlify-recaptcha netlify data-has-netlify=\"true\"") -}}
+            {{- safeHTMLAttr (print " data-has-netlify=\"true\" netlify") -}}
           {{- else -}}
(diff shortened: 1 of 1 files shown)


layouts/partials/fragments/contact.html
```diff
@@ -19,9 +19,9 @@
           class="contact"
           method="POST"
           name="{{- .Params.form_name | default (printf "contact-form-%s" .Name) }}"
-          id="{{ printf "contact-form-%s" .Name }}"
+          id="{{- .Params.form_name | default (printf "contact-form-%s" .Name) }}"
           {{- if .Params.netlify -}}
-            {{- safeHTMLAttr (print " netlify-recaptcha netlify data-has-netlify=\"true\"") -}}
+            {{- safeHTMLAttr (print " data-has-netlify=\"true\" netlify") -}}
```
mpourismaiel: Should the netlify be the last attribute for it to work? That's weird. Also doesn't this disable the netlify recaptcha?

mpourismaiel:
Fantastic. We can do the async form in a follow up PR.

layouts/partials/fragments/contact.html
stp-ip: Doesn't need to be last, just happened to be that way after my tests I think.

Nope it still has netlify recaptcha. The `netlify-recaptcha` config on the form itself is only used for custom provided recaptcha in combination with netlify. We don't support that yet.

stp-ip:
Ok let's do async in a separate one. Also think there is an open issue for it.

I'll try to fix the second captcha being injected issue in another PR as well.

PR #418, opened 2018-10-28
Make list fragment collapsible by a variable

### LINKED ISSUE #416: [list, toc] Nested content should be collapsible
feature

**What happened**:
ToC and List fragment can have nested content which would add much more data than what's actually needed for the user to see.

The nested content should be collapsible through a variable in the fragment's frontmatter.

Checkout `/docs` page of the demo (after #414 is merged).

mpourismaiel:
Adds new collapsible variable to list fragment (makes the content of article collapsible)
Makes ToC not render anything if it doesn't have any content

fixes #416 

**Release note**:
```release-note
- list: List fragment can now be collapsible
```


stp-ip:
Not expanding, when title is overflowing into next line.

stp-ip:
Should the icon change depending on the state aka between collapsed and non collapsed?

mpourismaiel:
@stp-ip Mostly we don't have animations in the page. I think the only animation that we have outside of bootstrap is for modal and portfolio's hover. Should I add it here?

stp-ip:
Mhh let's keep it for now. Still leaves the bug, when a title overflows.

mpourismaiel:
Do you mean the "Fragment implementation" article? I think the content in it has only one header and the code for ToC only render the ToC if it has more than one header. Will check why it renders the chevron.

mpourismaiel:
Fixed the issue. The problem was with the ToC fragment rendering without content.

PR #422, opened 2018-10-29
Make ToC fragment work standalone outside slots

### LINKED ISSUE #420: [toc] Fragment doesn't work outside of slots
bug

**What happened**:
ToC fragment only works in slots. The content variable inside it doesn't work.

For ToC fragment to work everywhere.

Try to use ToC fragment without slots.

The variable in the fragment only contains an string. It should contain the Page type instead of String type.

**Environment**:
- Syna Theme version:  
- Hugo version:  
- Others:

mpourismaiel:
Makes ToC fragment standalone and adds `/dev/toc` page to test the cases that ToC fragment supports.

fixes #420


stp-ip:
Still reviewing, but this breaks: https://deploy-preview-422--syna-demo.netlify.com/dev/empty-subitems/
Also the warning within the helper is tailored towards members only instead of more broadly phrased.

stp-ip:
Will take another fresh look at the code change tomorrow.

mpourismaiel:
That would be great. Also added the margin to the warning helper.

stp-ip:
Still breaks: https://deploy-preview-422--syna-demo.netlify.com/dev/empty-subitems/

stp-ip:
This breaks the collapsible functionality.

mpourismaiel:
Will switch to this after v0.14 fixes.
I check collapse in FAQ and navbar. Where does it break?
Will fix the warnings in empty subitems.

stp-ip:
The toc items seem not to support collapsible with the recent code changes in this PR. I might be wrong so. Will take another testing and review run at it tonight.

mpourismaiel:
Oh it's because of the conflict not being fixed. On it.

stp-ip:
Will do another test run tomorrow. Hopefully I can do some merges also. :)

PR #424, opened 2018-10-31
WIP: Add new payment fragment using Stripe

### LINKED ISSUE #9: Create payment fragment (stripe)
feature

**What happened**:
No payment fragment.

Payment fragment based on stripe.

Look for a payment fragment.

Also needs backend component hosted.

**Environment**:
- Syna Theme version:  v0.7.0
- Hugo version:  v0.30.2

mpourismaiel:
New `payment` fragment is added using Stripe.

fixes #9 

**DO NOT MERGE** This is the POC of payment fragment. The ajax request and the layout need to be discussed. The PR is ongoing.

**Todo:**

- [ ] Extra fields
- [x] Display price
- [ ] Display success/failure
- [ ] Switch to JSON requests

**Release note**:
```release-note
- payment: New `payment` fragment has been added, powered by Stripe
```


mpourismaiel:
Currently the result looks like this:

image

stp-ip:
From a design perspective I would say the price should be under the form input next to the submit button.
```
Price: 10€                      Submit
```

stp-ip:
Button default should be "Pay Now" I think.

stp-ip:
Also as a standalone payment fragment without pricing. The pricing might actually be a field, that can be used to set the price or can be overwritten by the pricing fragment. What do you think?

layouts/partials/fragments/payment.html
```diff
@@ -0,0 +1,86 @@
...
+  currency: "{{ .Params.currency | safeJS }}",
+  options: {
+    classes: {
+      base: 'text-body form-control row mx-0 px-2',
+      invalid: 'text-warning',
+      empty: 'text-body',
+      complete: 'text-success'
+    },
+    cardNumber: {
+      placeholder: '{{ .Params.payment.placeholder | default "4242 4242 4242 4242" }}'
```
stp-ip: placeholder isn't being used or?

mpourismaiel:
> From a design perspective I would say the price should be under the form input next to the submit button.
> 
> ```
> Price: 10€                      Submit
> ```

The feature is still not implemented completely, but in place of price in you example I added the failure/success message. I think we can add price here and move the message after under the button. Is that a good solution?

> Button default should be "Pay Now" I think.

Alright. Sounds good.

> Also as a standalone payment fragment without pricing. The pricing might actually be a field, that can be used to set the price or can be overwritten by the pricing fragment. What do you think?
> 
> So various modes of operation:
> 
> * default value set in config
> * pricing overwrites default
> * user overwrites default (input field) aka donation
> 
> This might warrant a config value such as "user_defined = true" or so. It's false by default and would grey out the field or not use a input field for price, but when defined the user can overwrite the default value and values set by pricing.

I was thinking of adding a pricing section to the payment fragment. Something like tabs on top of the form, clicking on each pill would change the price.
The user input seems like a good idea. Does it need a minimum field as well? Something like `donation_field/user_defined: boolean` and `minimum_price: number`. Combination of those fields would add a validation to the input. How does that sound?

stp-ip:
Instead of using minimal price we could use price as the minimal one. So if no price is configured there is no minimum. With a price configured it's automatically the minimum, but can be changed be the user.
Would be one less variable.

Also wouldn't structure the boolean under donation. Could be other things too, donation just seemed like the best example.

For the message I think under the button should be fine. Will need to check on a live interaction again.

stp-ip:
Also as always, if you feel I might miss something or you want to do something differently, don't hesitate to bring it up.

stp-ip:
Added the test account to get this in testing mode. \o/

stp-ip:
Adding additional metadata might be a good idea:
* [ ] page it was triggered from aka "about.example.com/donation/"
* [ ] product name

Product name might end up being added separately later as a group or separate key pair, but feels like a good initial thing to add or keep track of.

stp-ip:
Also having issues with CORS not allowed in firefox. Will see, if I can debug.

layouts/partials/fragments/payment.html
stp-ip: Not used it seems so we could remove this.

stp-ip:
Also the various ideas of pricing fragment interaction, default price and user changeable price still stands.

stp-ip:
Another big thing that I would love to see is recurring charges. That would probably fall under billing in stripe terms. Need to look into that more.

stp-ip:
Also for the simple case sending out receipts needs to provide receipt_email, which is still missing I think.

stp-ip:
Relevant for european users is sales tax support. In the initial PoC we just charge a set amount via stripe.

With that in mind we have various options to handle taxes. Locally and then just use stripe for charges.
That means that providing a VAT receipt needs to be done on our side too.

stp-ip:
Validation seems to have issues. Entered my email and the button still is greyed out. Reloading the page fixes this, but removing and reentering the email, will lead to a non functioning button again.

To not forget the meta data would be helpful:
* [ ]  page it was triggered from aka "about.example.com/donation/"
* [ ] product name

PR #426, opened 2018-11-03
Create new config fragment

### LINKED ISSUE #132: Support custom CSS and JS injections
feature

**What happened**:
We currently have the option to inject additional css and js and overwrite via Hugo. It's currently not possible to disable assets.

Currently working:
* css + js injection via config.toml config
* `exampleSite/static` overwrites the same file in `theme/static` <-- can be used to overwrite our files

Is this enough. Easy injection, more advanced replacement.

Related: #86

mpourismaiel:
Adds new config fragment and it's showcase/documentation page

fixes #132 

**Release note**:
```release-note
- config: New config fragment
```

### DIFF
--- a/exampleSite/content/fragments/config/code-config.md
+++ b/exampleSite/content/fragments/config/code-config.md
@@ -1,0 +1,28 @@
++++
+fragment = "content"
+weight = 110
++++
+
+```
++++
+fragment = "config"
+
+# Define meta tags
+[[meta]]
+  # html = "<meta meta charset="utf-8">"
+
+# Define link tags
+[[link]]
+  # block = false # Default is true
+  html = "<link rel='stylesheet' href='https://bootswatch.com/4/united/bootstrap.min.css'>"
+
+# Define scripts
+[[script]]
+  # block = true # Default is false
+  html = """
+  <script>
+    document.querySelector("#code-config").style.cursor = "url('http://wiki-devel.sugarlabs.org/images/e/e2/Arrow.cur'), auto";
+  </script>
+  """
++++
+```
--- a/exampleSite/content/fragments/config/code-config.md
+++ b/exampleSite/content/fragments/config/code-config.md
@@ -8,21 +8,18 @@ weight = 110
 +++
 fragment = "config"
 
-# Define meta tags
-[[meta]]
-  # html = "<meta meta charset="utf-8">"
+[[config]]
+  type = "css" # Acceptable values are icon, meta, link, css, js. Default is empty. Would not add anything on empty.
+  # block = true # If set to true, would inject the code to the <head> tag. Default is false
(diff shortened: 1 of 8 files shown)


mpourismaiel:
Will add support for injecting scripts and styles from nearby files.
Current approach allows for configuring meta and adding scripts and styles and images and etc. from other domains, mostly CDNs. I'll combine in with support for local files.
The support will likely work like image fallthrough. First I'll check for files in the fragment directory, then in the page directory and next in the static directory. Don't think linking to `_resources` directory would be possible though. I'll check that too. If it would be possible then the order would be:
`fragment directory > page directory > _resources directory > static directory`.
How's that?

mpourismaiel:
So if I understand correctly, you want to rename `file` to `resource`, which I definitely agree with. I wasn't really thinking about names. And we keep the `html` variable as they are.
Icon is used for favicon in `link` tags. It's possible to change favicon of different sections this way. Not really good to do but with file support it's easier to do. `html` is not something I'd want to endorse.
`config` fragment would mostly be in the `_global` directory I assume and scripts would mostly be in the `static` directory. Stuff in `assets` directory aren't used by Hugo unless they're accessed within the layout files so we would need to support them somehow but supporting them would mean we need to define a third variable since if the `resource.Get` function is called either the file is found or the build breaks, so wouldn't be able to fall through it. That's why I'm not able to implement resources in the config. The reason I didn't implement the fragment directory for files is that assets for pages would mostly be in the `static` directory. Will add fragment directory support to make it work like image fallthrough.

stp-ip:
Yes I meant renaming to `resource` makes sense.

Maybe it's good enough to support only the static directory and no page or fragment level. That makes our code easier and customization less error prone.

We could start without `html`. Basically `resource` taking a file or a URL. Would that be a better and cleaner first iteration?

stp-ip:
OK holding off on the merge until this is done. \o/

mpourismaiel:
Remove page level resources.

stp-ip:
So merge or wait until we simplify it into resource, which takes a file or url?

mpourismaiel:
I don't think there's anyway for resource to be in the config just yet. I think merge for now and create an issue for resource.

mpourismaiel:
PR #424 ([js partial]((https://github.com/okkur/syna/pull/424/files#diff-dd578a119f2073fb9e68d4e9d6aa377b)) adds the ability to specify which file would be loaded async and which loads synchronously. With that merged we can fix the issue globally without the need for merging files. This PR just fixes the issue for editor fragment, which is a built in fragment, not much overhead, predictable, which would make kinda sense to merge into the other bundle. Also reduces the amount of files we download separately in the page. We can later on bundle all the files together, with the js registry fully working.

stp-ip:
Ok merging this as is, but we should open the issue for v0.15 so that we don't release something just to change it within v0.16

PR #432, opened 2018-11-04
Remove .text-muted

### LINKED ISSUE #415: Use text-black and text-white for all text.
feature

**What happened**:
Currently we use text-dark and text-light.

Use text-black and text-white instead to decouple colorization of light and dark colors from impacting readability.

mpourismaiel:
Removes the instances where `text-muted` class from Bootstrap where being used.
One instance already used `text-color` helper. The two other instances where not using helper because of the `muted` class. Since that's not the case anymore, migrated them to use the helper.

fixes #415 

This partially fixes #415 since the other parts of the issue where previously fixed.

**Release note**:
```release-note
- BREAKING: Removed `.text-muted` from the theme. The class can still be used in custom layout files.
```


stp-ip:
```ERROR 2018/11/04 12:24:09 Error while rendering "page" in "docs/fragments/": template: _default/single.html:9:7: executing "_default/single.html" at <partial "helpers/fra...>: error calling partial: template: partials/helpers/fragments-renderer.html:18:8: executing "partials/helpers/fragments-renderer.html" at <partial (print "frag...>: error calling partial: template: partials/fragments/nav.html:175:3: executing "partials/fragments/nav.html" at <partial "helpers/sea...>: error calling partial: template: partials/helpers/search-result-template.html:15:12: executing "partials/helpers/search-result-template.html" at <partial "helpers/tex...>: error calling partial: template: partials/helpers/text-color.html:1:16: executing "partials/helpers/text-color.html" at <.self.Scratch.Get>: can't evaluate field Scratch in type interface {}
```
Build error ;)

stp-ip:
Also side question. Why are we using text-secondary in the footer?

mpourismaiel:
Stupid Hugo watch mode. Although this made me understand how it works so fast.
The copyright secondary color has been there for a long time. Should I change it?

stp-ip:
It sort of feels right, but secondary might be overwritten and end up being a wrong thing. How about text-black-50 on lighter backgrounds and text-white on darker ones. Does that work or is the contrast not good enough?

mpourismaiel:
That gives us this:

image

I kinda like it. Pushing the changes.

mpourismaiel:
We can change link color while we're at it. Any ideas what color we should use?

stp-ip:
Primary for general link color seems good, but for primary background we should probably use something differently. Not sure. Will think about it.

A few other thoughts.
Black text for white, light and secondary?
White text for dark and primary?
Link text primary for white, light (and secondary?)?
Link text something else for dark and primary?

stp-ip:
Maybe secondary for link text colors for dark and primary?

mpourismaiel:
We already have `body-50` on white, light, secondary, primary and `white` on everything else. I think that's already good. For link `primary` on white, light and `secondary` for everything else seems good.

Also I think we should provide a way to customize text colors. Maybe something like the following in `config.toml`:
```
  [params.text_colors]
    primary = "white"
    secondary = "white"
    dark = "white"
    light = "body-50"
    white = "body-50"
```

Maybe a way to customize link colors would be too much. We can use primary and secondary for dark and light backgrounds (not theme names).

stp-ip:
Customizing text color and link color would already modify quite a bit with our use of bootstrap.

Also I want to prevent using `secondary` for text as this might be colored. On the other hand both body-50 and white are not customizable so better for us to use.

Happy to hear the thought process on allowing more customizations. As especially with the config fragment these color changes can easily be done in a more advanced state. So my thinking is to keep the contrast relatively unchangeable and only be able to change some base colors as we do now and let the user with more needs use the config fragment.

stp-ip:
Still needs the last color changes as mentioned and a decision on the link-color idea.

stp-ip:
How about this:
dark background: link color = primary, hover link color = light
primary background: link color = light, hover link color = primary

Does that make sense as a first iteration?

stp-ip:
Any update?

mpourismaiel:
I'm working on events still. Will get on this after that if it's okay.

mpourismaiel:
> How about this:
> dark background: link color = primary, hover link color = light
> primary background: link color = light, hover link color = primary
> 
> Does that make sense as a first iteration?

On primary background, primary hover color would not be visible. Should I go with secondary instead?

mpourismaiel:
Looks like this with dark hover in primary background:

image

image

PR #436, opened 2018-11-06
Fix editor fragment not loading

### LINKED ISSUE #390: [react-portal,editor] React components don't always show up
bug

**What happened**:
Editor or other react portals don't always show up

To be sure that every react component is always rendered

Checkout `react-portal` or `editor` fragment examples. The issue happens occasionally.

**Environment**:
- Syna Theme version: v0.13
- Hugo version: v0.48
- Others:

mpourismaiel:
Merge `syna-editor.js` and `syna-react.js`. Doesn't cost much if editor fragment is not used but fixes the editor fragment not showing up quite easily.

fixes #390 

**Release note**:
```release-note
- editor: Fix fragment not showing up occasionally
```


stp-ip:
Doesn't that mean this issue would show up on other react usages too? We can't all merge into the react fragment.

mpourismaiel:
PR #424 (js partial) adds the ability to specify which file would be loaded async and which loads synchronously. With that merged we can fix the issue globally without the need for merging files. This PR just fixes the issue for editor fragment, which is a built in fragment, not much overhead, predictable, which would make kinda sense to merge into the other bundle. Also reduces the amount of files we download separately in the page. We can later on bundle all the files together, with the js registry fully working.

stp-ip:
Ok happy to merge as a fix for now and reevaluate later with the js registry fully working.

Yeah another merge \o/

PR #439, opened 2018-11-10
Update config fragment to accept resource instead of file

mpourismaiel:
Config fragment now accepts resource instead of file which would accept both url and file.


mpourismaiel:
No a relative path to the static directory only works. So if you have `static/file.js`, the value for `resource` will be `file.js`.

PR #444, opened 2018-11-26
Add stripe with events

### LINKED ISSUE #9: Create payment fragment (stripe)

### LINKED ISSUE #218: Add event fragment
feature

**What happened**:
Currently showcasing events is hard.

Make showcasing events easier.

* full event calendar also outputting JSON and webcal (specific page to make custom outputs easier?)
* fragment to showcase x next events being populated by JS from JSON

So we have the fragment with simple html + JS to populate from full JSON and we have the full event html, which uses JS to populate it from full JSON. Both link to a webcal link.

mpourismaiel:

fixes #9 
fixes #218 

I closed the payment PR since this is based on that. Keeping the note on the previous one here:

To not forget the meta data would be helpful:
* [ ]  page it was triggered from aka "about.example.com/donation/"
* [ ] product name

**Release note**:
```release-note
- stripe: Added new stripe fragment
- events: Added pubsub to the theme. Fragments can now seamlessly communicate with each other
```


mpourismaiel:
Fixed the validation issues. Had to update the dependency.

Regarding the url, we can receive that from the request headers, can't we? I'll add the value to the request body as well.

Regarding the product however, I'm not really sure where to add it. Do you have any ideas?

stp-ip:
For the product I would use meta-data. Might not be supported within our function, haven't checked. If it isn't we can open an issue and look into it in another PR.

mpourismaiel:
I send the title of the pricing plan if a plan is selected. The product is set to empty by default in the stripe fragment. We don't make use of it in the function, only log it so when we publish the function it's shown in there.

stp-ip:
Before merge let's cleanup the conflicts first ;)

mpourismaiel:
Resolved conflicts and added small documentation.

PR #449, opened 2018-12-03
Add support for labels in portfolio fragment

### LINKED ISSUE #225: [portfolio] overlay tags
feature

**What happened**:
Currently one can't add additional information to a portfolio item.

Additional information for portfolio items such as "design", "exit", "client" etc.

mpourismaiel:
Add support for labels in the portfolio fragment

fixes #225 

This is a PoC. I didn't follow the specs you provided, instead went with the icon version of the asset object. Let me know if you don't like it.

**Release note**:
```release-note
- portfolio: Add support for labels
```


stp-ip:
One thing this does make possible is filtering by tag/label. If we would reuse tags, we could make this possible, but on the other hand it feels unnecessary. What do you think?

mpourismaiel:
It would be possible but I'm not really sure how useful would that be.

stp-ip:
Let's go with this for now.
Merging after the conflicts are resolved.

PR #465, opened 2018-12-26
Add multiple prices support to stripe fragment

### LINKED ISSUE #459: [stripe] Add ability to enter amount within payment fragment
feature

**What happened**:
Currently one can add a default value or use events to change the amount.

Add ability to customize the amount within the payment fragment, when enabled. Disabled by default.

### LINKED ISSUE #479: [Stripe] Price normalization doesn't work with currencies with more or less floating points
bug

**What happened**:
Currencies like Yen that have more or less floating points (for example Rial or Yen don't have any floating points) don't get calculated properly.

For Stripe fragment to support all the currencies supported by Stripe.

Use Yen for currency in Stripe fragment and submit the form.

We need a much better price parsing strategy.

mpourismaiel:
Add support for multiple price values and editable price in stripe fragment

fixes #459 
fixes #479


stp-ip:
Also: No placeholder present so users might have a hard time to know, what to input.

stp-ip:
A few things:

* dropdown is too big as the biggest size would be 3 chars we can make this align with the dropdown button I think
* dropdown not overlayed (as you mentioned)
* price parsing doesn't seem to work perfectly aka "50,00" is considered not a number (is the dependency not as generic as I thought?)
* price parsing allows "50", "50.00", but not "50.55", "50,00" etc.
* using the increment up and down button switches the price format from "20.00" to "21", which feels weird
* breaks our event fragment event it seems
* mobile view breaks alignment of submit button (see screenshot above)

First review. Will take another look tomorrow with testing through the various variations such as single currency, multi currency with choices, but no custom price etc.

We are getting closer to something I feel like we can put out for feedback. Awesome start into 2019.

stp-ip:
Also the code examples are not updated yet, but I assume that's intentional to not duplicate work that can be done after we agree on the implementation.

mpourismaiel:
The price issues were related to the `input[type=number]` I was using. Changing it to `input[type=text]` fixes the issues but allows for not number entries. I tried adding the validator to the price input but it causes some problems that would take more time.

stp-ip:
* [ ] currency dropdown resizes depending on selection (example: $ vs CA$). Should be fixed size and aligned with dropdown field.
* [ ] site won't build without preselected [[prices]]. It should function with custom (allow_change), multi choice, event, default price and the combination of multi choice and allow_change.

assets/js/stripe.js
```diff
@@ -28,11 +29,20 @@ function onSubmit(configId, form, stripe, card) {
...
+        let price = serializedForm.price_text;
+        if (serializedForm.custom_value === "true") {
+          price = serializedForm.custom_price_text;
+          serializedForm.currency = form.$('[data-input=currency]').attr('data-value');
+        } else if (serializedForm.multichoice === "true") {
+          serializedForm.currency = form.$('input[name=price_text]:checked').attr('data-currency');
+        }
+        delete serializedForm.custom_price_text;
+        serializedForm.price_text = price;
+        serializedForm.price = parsePrice(price) * 100;
```
stp-ip: Is that really always the case? Even for Japanese Yen for example?

stp-ip:
* [ ] open separate issue for validator stuff related to pricing input

assets/js/stripe.js
mpourismaiel: No not really but I'm at a loss with normalization. `parse-price` returns float if the entered value looks like a float and returns int when the value doesn't have a comma or period inside it. Also it doesn't take currency into account. It just takes out the numbers and converts comma and period into period, then casting the value to `Number`. I'm looking into other options for normalization.

exampleSite/content/dev/events/payment/stripe.md
```diff
@@ -5,9 +5,10 @@ weight = 120
 post_url = "https://us-central1-syna-222118.cloudfunctions.net/function-1/charge"
 stripe_token = "pk_test_36PckiAlsGm9KmHj9b034GAW"
 
-price_text = "20.00$"
-price = "2000"
-currency = "usd"
+[[prices]]
+  text = "20.00$"
+  amount = "2000"
```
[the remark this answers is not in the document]
mpourismaiel: Yeah I missed this one. Thanks.

stp-ip:
Went through the various edge cases again and tried to break as much as possible. Still open issues right now that I found:
* [ ] open separate issue for validator stuff related to pricing input
* [ ] open separate issue to iterate on price normalization
* [ ] test with currencies without decimal points such as YEN
* [ ] allow_change (no preselected prices) fails with no currency error [/fragments/stripe/]
* [ ] charges fail on /dev/events/payment with no source/customer (still looking into it)
* [ ] events should overwrite preselected, default and custom (?without being changeable?)
* [ ] allow_change = false doesn't take effect

mpourismaiel:
- Currencies with more or less floating points fail. We need to change price parsing strategy. Not sure how to do it right now.
- I'm removing `allow_change` since it doesn't really do anything from configuration perspective. `user_input` defines an input that can appear to the user. `allow_change` then would mean that `user_input` is there but is read only or something. Removing it makes more sense to me. Let me know what you think.
- I'm afraid I don't understand "events should overwrite preselected, default and custom".

stp-ip:
What I mean is that an event doesn't overwrite anything. Not with a preselected price and not a custom price.

mpourismaiel:
Okay. So I changed some stuff and it appears to be working but the change is quite extensive.

- The stripe fragment didn't work in the `/dev/events/payment` because contact fragment overrode the submit event of stripe fragment.
- Stripe update event now fully overrides the price but also adds a reset button in front of the event.
- There is a new type of event listener in `jq-helpers.js` that mimics the event delegation behavior of jquery.

Let me know what you think. I tried to test as much as possible but I have to dash for now. If you find anything please let me know.

assets/js/stripe.js
```diff
@@ -148,3 +148,169 @@ function updateStripeFragments(product, price, currency) {
...
+  WST: 100,
+  XAF: 100,
+  XCD: 100,
+  XOF: 100,
+  XPF: 100,
+  YER: 100,
+  ZAR: 100,
+  ZMW: 100,
+  ZWB: 100,
+};
```
stp-ip: Zero decimal currencies according to stripe (not a perfect list) https://stripe.com/docs/currencies#presentment-currencies:
mga, bif, clp, pyg, djf, rwf, gnf, ugx, jpy, vnd, vuv, xaf, kmf, krw, xof, xpf

Let's fix these to 1 and reiterate with a better list later.

stp-ip: Besides this it's a LGTM. All other things I'll create issues for.

mpourismaiel:
@stp-ip Changed the currency list to match the data you provided. Also merged master into this branch so the conflicts are fixed and nothing seems to be amiss. I'll fix the reported issues on top of master since this branch has gotten really big.

PR #467, opened 2019-01-02
Add scroll spy to content fragment

### LINKED ISSUE #443: [toc] Show current opened page/section within toc
feature

**What happened**:
Currently switching pages will start with a closed/collapsed toc.

When a new page and section is linked and accessed the toc should be uncollapsed and showing the current section one is browsing.

mpourismaiel:
- Add scroll spy to content fragment
- Make sidebar in content fragment sticky
- Open current page in toc in content fragment's sidebar

fixes #443


stp-ip:
This works in most cases, but not in all. It might need more thinking or documentation on how not to misuse.

A few things:
* `enable = true` is replaced by `sticky = true/false` and therefore needs to be set?
* `content` still used for toc? Feels weird to say `content_2.md` and also doesn't seem to take effect.

exampleSite/content/fragments/content/content-sidebar-right.md
```diff
@@ -12,6 +12,7 @@ subtitle = "Split in two!"
 [sidebar]
   title = "Sidebar"
   align = "right"
+  sticky = false
```
stp-ip: Couldn't sticky be "false" by default and we only need to set `sticky = true` if needed?

stp-ip:
* [ ] mark current/active page in toc sidebar (currently we only mark the specific section as active)

exampleSite/content/fragments/content/content-sidebar-right.md
mpourismaiel: It is false by default. I just added it in the examples to show off it's existence. I'll replace it with `#sticky = true # Default is false` to make it more clear.

stp-ip:
Happy to merge this as is and iterate on the toc for multiple content files in a follow up PR.

PR #475, opened 2019-01-03
Fix collapsible items in list fragment malfunction

### LINKED ISSUE #472: [toc] List uncollapses on click of subitem on another page
bug

**What happened**:
Switching pages via a subitem from a collapsible toc item uncollapses the item before switching pages. (see reproduction steps)

Don't uncollapse on switching sides as it is unnecessary and creates confusion for the user.

On https://syna-demo.okkur.io/docs/ open the collapsible fragment list and then click on any of the subheaders/subitems. It will first close the collapsible list and then redirect to the /fragments/page.

mpourismaiel:
Collapsible items in list fragment were entirely clickable, making them behave somewhat abnormally. Reducing the size of clickable area to only the title of each article (in list fragment) and ignoring clicks on the link seems to have cleared the issue.

fixes #472


stp-ip:
Awesome. Thanks for diving deeper and finding the actual problem.

PR #486, opened 2019-01-11
Patch some usability issues in stripe fragment

### LINKED ISSUE #482: [stripe] icon and button accessibility
feature

**What happened**:
For icons/buttons such as cancel event overwrite and edit with custom value no alternative text is given. Maybe also add a tooltip on hover?

We should be inclusive here and add a default text with translation possibilities.

### LINKED ISSUE #483: [stripe] icon to cancel value overwrite from event iteration
feature

**What happened**:
When an event overwrite happens we add an icon, which is hard to grasp.

Reduce the options for the user.

Do we really want the user to switch back? An even is there for a reason and most likely should overwrite the value. Making the user aware of other options feels problematic.

My view would be to not let the user know that he can switch back.

An event is most likely intentional and more intentional as any default values. Switching back, just adds another option, which reduces the likelihood of the payment being made.

### LINKED ISSUE #484: [stripe] simplify config and user interaction
feature

**What happened**:
Currently the various options seem to lack a consistent visual idea.

Group options visually into custom/input, event (plain) and preselected (button)

# idea
### custom input
```
[user_input]
  default = "20.00"
  currencies = ['usd', 'eur', 'cad'] # First currency will be used as default
```

### multiple choice
```
[user_input]
  default = "20.00"
  currencies = ['usd', 'eur', 'cad'] # First currency will be used as default

[[prices]]
  text = "30.00$"
  currency = "usd"

[[prices]]
  text = "50.00$"
  currency = "usd"
```

### single choice
```
[[prices]]
  text = "50.00$"
  currency = "usd"
```

### multiple choice
```
[[prices]]
  text = "30.00$"
  currency = "usd"

[[prices]]
  text = "50.00$"
  currency = "usd"
```

This would mean we only have 3 visual modes:
* event induced aka plain number
* selection aka button (with possible change)
* input box

And 3 showcases on the stripe page:
* single button (no change)
* multiple buttons (change)
* input box

What do you think?

mpourismaiel:
- Add tooltips to icons
- Remove cancel button from event invoked price
- Change view of the fragment for different configurations

fixes #482 
fixes #483 
fixes #484


stp-ip:
Based on this I did a few design iterations #494

stp-ip:
Product not updated within `content/dev/event/payment/pricing/plan1.md` 
Should be: `button_url = "/dev/events/payment?event=pricing:change&product=Starting plan&price=$9.99/mo&currency=usd"`

stp-ip:
I think we might need to print a note for the configuring single [[prices]] with [user_input]. As this would overwrite the single price. Maybe we should fail hard here? Allow one or the other and only allow both, when [[prices]] contains more than one element?

mpourismaiel:
I think this is now ready for another review.

PR #528, opened 2019-01-29
Add fontawesome pro support and document fontawesome behavior

### LINKED ISSUE #526: [fontawesome] Document usage of Pro version
feature

**What happened**:
Currently we sorta hardcode fontawesome in.

Make it easier and document how to use the pro version from fontawesome.

Good preparation for the coming subsetter I think.

mpourismaiel:
- Add fontawesome pro support
- Document fontawesome behavior

fixes #526 

**Release note**:
```release-note
- Fontawesome: Add Pro support
- Documentation: Add documentation for FontAwesome
```

### DIFF
--- a/assets/styles/index.scss
+++ b/assets/styles/index.scss
@@ -15,12 +15,17 @@ $gray-200:            #e9ecef !default;
 $border-color:        $gray-200;
 $list-inline-padding: 5px;
 $dropdown-item-padding-x: 1rem;
-{{ if not .Site.Params.fontawesome.disabled }}
-@import './fontawesome/fontawesome.scss';
-@import './fontawesome/regular.scss';
-@import './fontawesome/solid.scss';
-@import './fontawesome/brands.scss';
-{{ end }}
+{{- with .Site.Params.fontawesome -}}
+  {{- if not .disabled }}
+    @import './fontawesome/fontawesome.scss';
+    @import './fontawesome/regular.scss';
+    @import './fontawesome/solid.scss';
+    {{- if .pro }}
+      @import './fontawesome/light.scss';
+    {{- end }}
+    @import './fontawesome/brands.scss';
+  {{- end -}}
+{{- end }}
 @import './bootstrap/bootstrap.scss';
 @import './helpers';
 @import './header';
--- a/assets/styles/index.scss
+++ b/assets/styles/index.scss
@@ -15,16 +15,14 @@ $gray-200:            #e9ecef !default;
 $border-color:        $gray-200;
 $list-inline-padding: 5px;
 $dropdown-item-padding-x: 1rem;
-{{- with .Site.Params.fontawesome -}}
-  {{- if not .disabled }}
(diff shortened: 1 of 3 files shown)


exampleSite/content/docs/development/content.md
```diff
@@ -25,6 +25,18 @@ You can change these colors by editing them in `config.toml`.
...
 Syna customizes some parts of the theme via custom css, which is available in the `assets/styles` directory.
 
+### Fonts
+
+FontAwesome Free is supported by default and is used throughout built-in fragments. You can disable it by setting `fontawesome.disabled` in your `config.toml` to `true`.
+
+If you want to add FontAwesome Pro to the theme, the following recipe may be helpful.
+
+- Move all the files from `fontawesome-pro-[version]-web/webfonts/` (from the archive) to `static/fonts` directory in the theme.
+- Move all the files from `fontawesome-pro-[version]-web/scss/` (from the archive) to `assets/styles/fontawesome` directory in the theme.
```
stp-ip: Does this have to be within the theme? Wouldn't it be enough adding it in the website so it overwrites the theme files?

assets/styles/index.scss
```diff
@@ -15,12 +15,17 @@ $gray-200:            #e9ecef !default;
 $border-color:        $gray-200;
 $list-inline-padding: 5px;
 $dropdown-item-padding-x: 1rem;
-{{ if not .Site.Params.fontawesome.disabled }}
-@import './fontawesome/fontawesome.scss';
-@import './fontawesome/regular.scss';
-@import './fontawesome/solid.scss';
-@import './fontawesome/brands.scss';
-{{ end }}
+{{- with .Site.Params.fontawesome -}}
```
stp-ip: This breaks legacy config as we didn't check for the fontawesome config.

So it goes from working by default to it not working by default.

exampleSite/content/docs/development/content.md
stp-ip: Finally had time to test and works as expected within the site. So we should definitely recommend only overwriting this within the site not the theme.

assets/styles/index.scss
stp-ip: Do you feel we are early enough to go for cleaner code and break for v0.15? I'm fine with either.

stp-ip: Additional point. Either way I think the config should be easier.

fontawesome not configured (minimal config) -> free works
fontawesome disabled -> don't include files
fontawesome + pro -> include pro files

With this we would at the same time support the old way of doing things. So probably not breaking at all.

mpourismaiel: I'm fine with cleaning up the code across the theme as soon as possible. Because of our plans for v0.15 I think cleaning up before releasing makes sense.

exampleSite/content/docs/development/content.md
mpourismaiel: Sorry I though `assets` directory didn't fall through as `static` directory does.

PR #551, opened 2019-03-19
Add ability to change image size to specific aspect ratio

### LINKED ISSUE #35: Specific container sizes on the website to prevent layout issues
feature

**What happened**:
Currently the image sizes for logos etc. are only recommended and might break the layout.

Set a height and max-height or so to keep the layout from breaking.

mpourismaiel:
Adds `resize` variable to `image.html` helper partial which will cause a resize on images to the set value. This is used in fragments with images to keep the image from breaking the layout if it's too big. There is a new `/dev` section (`/dev/image-sizes`) showcasing the results for images that are too big.

fixes #35 

**Release note**:
```release-note
- Image resize: Images are resized automatically to keep the layout in good shape when an image is provided that is too big 
```


stp-ip:
I wanna do a more thorough review with some testing so might take a bit.

layouts/partials/helpers/image.html
```diff
@@ -1,24 +1,43 @@
 {{/* Global resource fallback */}}
 {{- $image := .asset.image -}}
+
+{{/* Resource is used when file is not imported from static directory, in which
```
stp-ip: Not sure what you want to say here. Might just be me, but maybe could be described in more detail.

mpourismaiel:
I'm trying to change the sizes to a more simple version but it's not really easy. The ones that define only width or height (`x200` for example) are there because the other aspect of the image doesn't break the layout and maybe the user wants a really long or wide image and it wouldn't break anything. But for others since the size is bound from both css and html, we mostly need that specific size to appear. And since Hugo only fills, resizes and fits the image to the given dimension instead of using aspect ratio, we can't really do much except for changing the design a bit and use simpler numbers. For example `508x381` would become `500x375`.

mpourismaiel:
Also I worked a bit on allowing the user to disable the resizing and I can't think of any solution that would allow us to do something like this:

```
fragment resize: nil, page resize: nil, site resize: nil -> resize
fragment resize: nil, page resize: false, site resize: nil -> don't resize
fragment resize: true, page resize: false, site resize: nil -> resize
fragment resize: false, page resize: true, site resize: true -> don't resize
```

I can't think of a condition to write for this. I imagine we would use a variable to override the functionality like `resize_assets` and if it's set to `false`, we wouldn't resize the image but if it's not provided or if it's set to true, we would resize the image. But then if we want to disable the functionality on different levels, then we also need to provide a way to enable it on a lower level. For example the user might want to disable the feature globally or on a specific page but wants to activate it for a certain fragment.

It's too hard for me to code :D I tried using `or` function from Hugo's go template function arsenal but it doesn't really work well here. I also tried the codes from `chmod` function on Linux and can't figure out how to change that to fit here.

stp-ip:
In general I think forcing a specific size both width and height makes our design more predicable, if they really want to add something that is really narrow or such they can always resize the image themselves with a transparent background etc.
So my view is let's use specific sizes instead of just ratios. Using simpler values might help our documentation. Therefore probably a nice to do.

Additionally we should document automatic resizing on our user docs including the recommendation to use page or fragment based images as anything in static/ isn't going to be resized.

On the note of disabling resize.
We could do a global flag or document to use static/ as a first iteration. This is given anyway so documenting it as a solution to disable specific images from being resized seems to be ok. Afterwards we can listen for feedback.

What do you think?

mpourismaiel:
I changed the values to simpler once and removed one sided resizing(?) for item and hero images since `.Fill` doesn't support it. The images look good but for a really tall image the layout doesn't really look good. As you said the user should crop these ones themselves. But please checkout what it looks like and keep in mind that I agree that the user should fix these for themselves. The layout doesn't break for these fragments, it's just too long.

exampleSite/content/docs/global-variables/content.md
```diff
@@ -61,6 +61,10 @@ Any fragment that uses this variable type would show either an image or an icon.
 
 The type is introduced to make configuring images and icons same between different fragments.
 
+Some fragments support resizing the images provided to them using the asset object. These fragment document the size that your image will be resized to.
```
stp-ip:
```suggestion
Some fragments automatically resize images to better fit the layout. For the best result use the documented size of each fragment's asset object.
```

exampleSite/content/docs/global-variables/content.md
```diff
@@ -61,6 +61,10 @@ Any fragment that uses this variable type would show either an image or an icon.
 
 The type is introduced to make configuring images and icons same between different fragments.
 
+Some fragments support resizing the images provided to them using the asset object. These fragment document the size that your image will be resized to. 
+Please note that images inside the `static/` directory will not be resized. So if you want to disable the image resizing, put the image in that directory.
```
stp-ip:
```suggestion
*To disable image resizing use assets stored within `static/` as these are excluded from auto resizing.*
Please make sure to provide the correct size as documented for each fragment's asset object.
```

exampleSite/content/docs/global-variables/content.md
```diff
@@ -61,6 +61,10 @@ Any fragment that uses this variable type would show either an image or an icon.
...
+If not, you can move the image to page or fragment directory.
```
stp-ip: I think that is already added within the above paragraphs.

exampleSite/content/dev/image-sizes/vertical/item/item-left.md
```diff
@@ -0,0 +1,15 @@
++++
+fragment = "item"
+#disabled = false
+date = "2017-10-04"
+weight = 335
+background = "secondary"
+align = "right"
+
+title = "item - right"
```
stp-ip: mismatch between title and file name aka `left` vs `right`

exampleSite/content/dev/image-sizes/horizontal/item/item-left.md
```diff
@@ -0,0 +1,15 @@
++++
+fragment = "item"
+#disabled = false
+date = "2017-10-04"
+weight = 335
+background = "secondary"
+align = "right"
+
+title = "item - right"
```
stp-ip: mismatch between title and file name aka left vs right

PR #556, opened 2019-04-12
Prototype: Create new grid fragment

### LINKED ISSUE #545: [sidebar] Sticky navigation with split content files
**What happened**:
We currently make sidebar navbars sticky within one content fragment.

Maybe make the sidebar sticky for all content fragments?

https://about.txtdirect.org/docs/examples/ <-- multiple content files
https://about.txtdirect.org/docs/specification/ <-- one content file

mpourismaiel:
Create new grid fragment

I realize there are no issues related to this PR, I just thought it would be a good new feature for a more advanced case.

To create this I had to split the `slot` helper to work more closely to how `fragments` helper works. It now just detects how deep the slot is and if it's below the `nested_level` variable, collects the fragments that are going to be rendered there. Then the new `slot-renderer` helper renders the fragments.

This is done to allow wrapping each fragment in the slot with a custom template. I use this to wrap the fragments with a `div.col-...`.

This can be used to fix #545 

If you think this is unnecessary or needs enhancements or is confusing, let me know. If not, it needs documentation and we can provide simpler layout fragments using this one later on.

**Release note**:
```release-note
- Grid: New grid fragment
```

### DIFF
--- a/exampleSite/content/fragments/grid/config.md
+++ b/exampleSite/content/fragments/grid/config.md
@@ -1,0 +1,19 @@
++++
+fragment = "config"
+
+[[config]]
+  type = "css"
+  html = """
+<style>
+.bg-custom-column {
+  background-color: rgba(86,61,124,.15);
+  border: 1px solid rgba(86,61,124,.2);
+}
+
+.bg-custom-column h2 {
+  color: black;
+  color: var(--dark);
+}
+</style>
+"""
++++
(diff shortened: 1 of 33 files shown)


stp-ip:
On the edge about this. It adds a lot of complexity and user variance, that will in my view lead to more configuration issues and misuse issues. Having our opinionated fragments and the ability for any user to create their own custom fragments already allows huge flexibility.

Happy to revisit this, but would say we shelf this for now.

Let me know, if you think differently.

PR #558, opened 2019-05-04
Update stripe fragment's design

### LINKED ISSUE #494: [stripe] Design iterations
feature

**What happened**:
Currently we haven't optimized for an overall design for the stripe fragment.

Consistent and Syna style (bootstrap) stripe design.

Doing various iterations on design.

General:
`product` as text
`price` on the button

## event chosen price and single preselected price
screenshot_2019-01-12 pricing and stripe middot okkur - syna 2

## multiple preselected prices
screenshot_2019-01-12 pricing and stripe middot okkur - syna 3

## multiple preselected prices + change
screenshot_2019-01-12 pricing and stripe middot okkur - syna 6

## custom price via input field
screenshot_2019-01-12 pricing and stripe middot okkur - syna 7

mpourismaiel:
Updated stripe fragment's submit button to a more informative version introduced in #494 

fixes #494


stp-ip:
Overall I think this is already great. Makes the button much clearer. A few things that came up and might be good to discuss:
* single price still being shown, but could be removed due to duplicate information in button
* customizing button text is harder now, but I think this is fine for now

mpourismaiel:
Would you please elaborate the points? I don't really understand.

stp-ip:
Point 2 is just a comment so nothing to do. Adding the price makes the button less flexible as it can't be "Do something" as it would always say: "Do something $20.00". But it's a payments fragment so it is a reasonable compromise in my view.

Point 1 is about the duplicate information. The single price setting of the fragment now shows a price + the button, which includes the price. In my view this is duplicate information and should be reduce. See the following image:
Screenshot_2019-05-07 Stripe middot; Okkur - Syna

mpourismaiel:
This should fix it. It looks much cleaner this way.

image

PR #560, opened 2019-05-07
Fix global fragments not working on every section

mpourismaiel:
Replaces old variable with a more complete version of it, fixing section globals not working on the section level when the page is deeper than 2 sections.


stp-ip:
Merged into master via https://github.com/okkur/syna/commit/41d9dcb77b661c50e478a16844581650caffb7cf, but Github didn't pick it up.

PR #565, opened 2019-05-10
WIP: Implement table fragment using markdown

### LINKED ISSUE #399: [Discussion] Deprecate table
Table fragment and table support in item fragment can be removed in favor of markdown table. One thing that would be missed is hidden columns in smaller devices and another thing would be alignment. What do you think?

mpourismaiel:
Prototype of what a markdown table fragment would look like.

This implementation mimics the design and behavior of the frontmatter table fragment through the usage of shortcodes and search and replace through the html code of the markdown table. It's a really rough implementation right now but it's functional and gives a good idea of how this idea would proceed.

The good thing about this implementation is **1)** this code can be moved to the table helper and every fragment using `.Content` can make use of it and **2)** shortcodes are here. Using shortcodes in table has the advantage of allowing the user to use different data types along side each other, for example a button along an icon. Also **3)** the old implementation of table fragment (table helper, frontmatter table) doesn't need to be removed. 

If you have any suggestions please let me know.

- [x] table fragment
- [ ] table helper
- [ ] item fragment
- [ ] content fragment

fixes #399 

**Release note**:

Only table fragment is implemented right now.

```release-note
- table, content, item: 
```


mpourismaiel:
The syna table shortcode is hard to pull off. I managed the css only table with a bit of hacking around, doing the syna shortcode is harder though. I've experimented a bit with partials and shortcodes and did a bit of hacking with them and have an idea on how to do this but I think I'm overthinking it. I'll try to work on it again tonight. If you have any ideas on this please let me know.

stp-ip:
Anything that is still WIP here? Or good for a final review?

stp-ip:
Ping ;)

mpourismaiel:
I forgot about this PR. It's in WIP because I can't prevent the shortcode from rendering the code block and it's problematic. We can merge the PR and just go with it but I rather use the frontmatter table for now instead of messing the demo.

mpourismaiel:
The new markdown compiler might help. Let me go through it.

stp-ip:
Now with the ability to create our own inline shortcodes could we come up with a better table usage?
Keeping it open as our current table fragment is not good. We should deprecate it and use markdown tables, that are well designed.

orchid00:
this will be very helpful @stp-ip would it be implemented?

PR #570, opened 2019-05-27
Patch a few issues

### LINKED ISSUE #360: Move the minimal CSS to the theme side
With #319 we have a minimal CSS which is added directly to the HTML in order to have the basic layout of the page if the user doesn't receive the complete CSS. But the minimal CSS needs to be located in the user space since `readFile` function (defined by Hugo) only works in user space and doesn't read any files inside the theme (unless the path is hardcoded which is not reliable).

Track https://github.com/gohugoio/hugo/issues/4842 and whenever it's done or a new feature from Hugo is available which allows us to read a CSS file's content inside the layouts, move the minimal CSS to the theme.

mpourismaiel:
- Add the new auto generated files (have been missed in a previous commit)
- Move `syna-grid.css` from user side to theme side (bad approach, removes it from user side at least)
- Fix the `.Dir` deprecation warning, I have no idea how these are popping up

fixes #360


stp-ip:
Merged into master as these were fixes. Will cherry-pick these to v0.15 in a bit.

PR #573, opened 2019-06-12
Portuguese translation

eriicf:
Portuguese translation. I've just updated by comparing to the original English en.yaml


stp-ip:
Awesome thanks for addition.

Would it be ok to ping you before a release to get this updated, if necessary?

eriicf:
You can contact me, I'm happy to help.

PR #576, opened 2019-06-13
Fix linking in docs

stp-ip:
Migrate linking to use Hugo style references.

fixes okkur/website#7

**Release note**:

```release-note

```


stp-ip:
This currently breaks with our `/dev` content. We probably need to prefix all content in their with `dev-` or similar.

Happy for other solutions. On that note. Do have have the generation script checked into code? @mpourismaiel

mpourismaiel:
This looks great. Thanks a lot.

mpourismaiel:
There seems to be a problem with the build though.

stp-ip:
Yes see comment above. This is due to our /dev/ sub dir.

stp-ip:
This needs a solution as manually fixing this feels problematic and it keeps the live site unsynced from the Syna code.

stp-ip:
Any ideas? Happy to take another look at an idea next week.

mpourismaiel:
I don't really have an idea. I'm migrating to v0.60 so we might be able to fix it in there if there's a hidden solution somewhere.

stp-ip:
Prefixing the dev pages with "dev-" would probably help make them not matched and therefore work around the duplicate warning.
What do you think?

stp-ip:
Merging this now. Voice your concern now or hold it for ever.

mpourismaiel:
I thought I commented on this one. There's no problem with adding a prefix for specificity's sake. lgtm.

PR #577, opened 2019-06-13
French translation

Gilrain:
**What this PR does / why we need it:**
French translation based on the english file.


stp-ip:
Awesome. Thanks for the contribution.

Would it be ok to ping you before a release to get this updated, if necessary?

Gilrain:
> Would it be ok to ping you before a release to get this updated, if necessary?

Yes, please do.

PR #590, opened 2019-08-13
Fix global fragments with subitems showing warning

### LINKED ISSUE #582: Global fragment not working with member fragment
bug

**What happened**:
I'm creating a site using Syna, according to the docs, i should be able to add a global fragment to a section of my site, for example, the blog, by creating a _global folder and putting there the needed fragments, but this doesn't seem to work with the "member" fragment.
I got this error:

`Currently there is no configured member content files present. To add new members add a content file within blog/_global/author`

inside content/blog/_global/author i already have a index.md and a author.md files

Get a member fragment in all the blog posts

1. Setup a syna-start repo
2. Decompress the attached blog folder inside content, which has the following structure:
```
content/blog
├── entry <- all the blog's entry resources
│   ├── content.md <- content fragment 
│   └── index.md
├── _global
│   ├── author
│   │   ├── author.md <- member info
│   │   └── index.md <- member fragment
│   └── index.md
├── _index
│   └── index.md <- list fragment
└── _index.md
```
[blog.zip](https://github.com/okkur/syna/files/3433570/blog.zip)
3. run `hugo server`
4. this results in a failed build with the following error:
` Currently there is no configured member content files present. To add new members add a content file within blog/_global/author`

There is a way to make this work, but its very annoying, i'll try to explain it the best i can
1. start `hugo server` before creating `content/blog/_global/author/index.md` or comment all its contents
2. Create a new folder `content/blog/entry/author` and put inside a `author.md` with all the info of the author and omit a `index.md`file, also the `content/blog/_global/author/author.md` file must be present otherwise this doesn't work, 
3. Uncomment o create the member fragment in `content/blog/_global/author/index.md`, save 2 or 3 times, the first time you save it Hugo will output the same error, later saves will work fine

I think this is a feature that avoids code duplication and i think my use case is not an edge use case.

Also i was wondering if there is another way to accomplish the same thing using slots, i have seen them been used in the code, but i couldn't find any documentation about them.

**Environment**:
- Syna Theme version:  v0.15.1 and v0.15
- Hugo version:  v0.55.6/extended linux/amd64
- Others:

mpourismaiel:
Global fragments with subitems (member, items, portfolio, faq) now display their subitems instead of "empty subitems" warning.

fixes #582 

**Release note**:
```release-note
- Fixed global fragments with subitems showing warning instead of items
```

### DIFF
--- a/layouts/partials/fragments/faq.html
+++ b/layouts/partials/fragments/faq.html
@@ -1,7 +1,8 @@
+{{- $items := partial "helpers/subitems.html" . -}}
+
 {{- $self := .self -}}
 {{- $bg := .Params.background | default "white" -}}
 {{- .page_scratch.Add "js" (dict "file" "syna-collapse.js") -}}
-{{- $items := where (.page.Resources.Match (printf "%s/*.md" .Name)) ".Name" "ne" $self.Name -}}
 
 {{- partial "helpers/container.html" (dict "start" true "in_slot" .in_slot "bg" $bg "Name" .Name "Params" .Params) -}}
   {{- partial "helpers/section-header.html" (dict "self" $self "bg" $bg "params" .Params) }}
--- a/layouts/partials/fragments/items.html
+++ b/layouts/partials/fragments/items.html
@@ -1,4 +1,5 @@
-{{- $items := where (.page.Resources.Match (printf "%s/*.md" .Name)) ".Name" "ne" .self.Name -}}
+{{- $items := partial "helpers/subitems.html" . -}}
+
 {{- $self := . -}}
 {{- $bg := .self.Scratch.Get "bg" }}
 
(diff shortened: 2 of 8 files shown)


stp-ip:
Using Hugo v0.56.3 I get an empty fragment for member, faq, items, pricing, hero, graph, editor and portfolio.

So this doesn't seem fixed. It just doesn't print the subitem error. Or am I missing something?

mpourismaiel:
We still have the subitem problem with 0.56.3. I still can't get it to work and the only solutions right now are `[page]/[fragment].md` files with subitems in `[page]/[fragment]/` directory but I didn't move the files around as this is most likely a hugo bug that needs to be fixed or there is a solution that I haven't found yet but either way I thought moving files around would be confusing. Test it with 0.55.6 and it should work or move the `[fragment]/index.md` file for now. That should show the fix for the global fragments.

stp-ip:
Ah so specific fix for the global fragments not generally subitems. Thanks for clarifying. Will check it out tomorrow.
Also filing a bug with Hugo is on my list ;)

PR #593, opened 2019-08-15
WIP: Add url_ref support to fragments that support url

### LINKED ISSUE #574: Add possibility to use "ref" for links within front-matter
feature

**What happened**:
Currently hugo reference style linking within buttons etc. isn't possible.

Enable "ref" style links even in front matter.

This should also make link breakage more noticeable.

mpourismaiel:
Add url_ref variable wherever we support url variable.

Still needs documentation and more tests.

fixes #574 

**Release note**:
```release-note
- Added new `url_ref` variable
```

### DIFF
--- a/layouts/partials/fragments/buttons.html
+++ b/layouts/partials/fragments/buttons.html
@@ -1,14 +1,14 @@
-{{- $self := .self -}}
+{{- $self := . -}}
 {{- $bg := .self.Scratch.Get "bg" }}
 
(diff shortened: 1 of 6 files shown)


stp-ip:
Let's hold this off for the moment. I feel adding another abstraction/config option doesn't make things easier.

stp-ip:
This needs upstream support or else it won't be a good user experience as far as I can see. Happy to revisit this so.

PR #618, opened 2019-10-27
Fix missing space before height and width attributes in asset img tag

pv2b:

The hero fragment allows a user to include an asset in the hero banner. However, the HTML for the img tag is broken if width and/or height is set:

    logo</img>

Notice missing space before width. This PR fixes this.

**Release note**:
```
Hero fragment: Added missing space before height/width attributes in <img> tag
```


stp-ip:
Awesome. Thanks for the contribution. Looking forward to even more :+1:

PR #628, opened 2019-11-04
Add first tests

### LINKED ISSUE #625: Add end to end tests
feature

**What happened**:
There are no tests beyond the manual tests we use to validate theme functionality.

Use a tool such as Cypress to create end to end tests.

mpourismaiel:
Create a few tests using cypress

fixes #625 

Adding Cypress caused an update on Bootstrap which didn't cause any issues with the theme but I'm gonna investigate it further. Also I'm gonna migrate a few other tests and add more tests for the functionality. Most tests that I've written test the core functionality not user interactions or styling. I'll add more tests in that area moving forward.

**Release note**:
```release-note
- We now have automated tests to keep from breaking things.
```


stp-ip:
For something like this splitting of the bootstrap update into its own commit would be great. Even so merging the update in with e2e tests feels problematic for future reference.

stp-ip:
Also the bootstrap update was more of a downgrade from v4.3.1 to v4.1.3, which does make this a merge blocker in general.

cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
```
stp-ip: Example command intentionally committed? Why?

mpourismaiel:
I had removed the cypress starter files. Can't remember what happened. They shouldn't be here. I'll remove them. Also the bootstrap trick is something else that cypress did. Didn't check the diff. Sorry.

mpourismaiel:
I resolved to force pushing this one. Removed useless files, reverted bootstrap downgrade, merged master into the branch and pushed it again.

PR #637, opened 2019-11-21
Remove stray end tag div and indent block

Marzal:
One less w3 validator's error without the stray div.

I don't know exactly what this block of code does, but anyway better to have it clean and tidy.
There is another error with `` been empty, but I'm not sure if is there for a especial use case.

**Release note**:
```release-note
Minor fix: remove "tray end tag div" error
```

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -11,22 +11,21 @@
 
     <div id="react"></div>
     <div class="modal fade" tabindex="-1" role="dialog" aria-hidden="true">
-        <div class="modal-dialog" role="document">
-          <div class="modal-content">
-            <div class="modal-header row mx-0">
-              <div class="modal-title col px-0">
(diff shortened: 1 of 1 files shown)


stp-ip:
This is used as a general component for modal views. Most likely this shouldn't live within the baseof.html. Looping in @mpourismaiel as he has a better grasp on this part of the code. 
I would say lgtm + separate cleanup pushing the modal code into its own file.

mpourismaiel:
The change looks great. Thanks a lot. The entire block is used in the `assets/js/modal.js` to display a modal for any fragment that needs one. The empty `img` tag is used for the same purpose. We can remove the block and move it to the `modal.js` script, but I think that needs another PR.

stp-ip:
Yeah moving the img to modal feels like a good solution as a follow on PR. Opening an issue for this now.

Besides this merging.

PR #639, opened 2019-11-30
FAQ: Use primary color instead of fixed value.

c-mauderer:

Currently the color of the FAQ title is set to a fixed `#00838f` which is the primary color for Syna. If the primary color is adapted, the fixed value doesn't match any more.

I noted that the background color of the FAQ items seems to be a default gray too. But I don't know where it is set or what value would be desirable to match as many color choices as possible.

**Release note**:

```release-note
Set FAQ question color to the primary color.
```


stp-ip:
This needs a rebuild of the static assets or else it's only within the scss files.

I think `make build` should generate everything necessary and then add that as well.

c-mauderer:
I'll try to generate the files. Sorry that I missed it. As an embedded developer web projects are not my usual type of projects.

stp-ip:
Happy to help, if you run into any issues.

c-mauderer:
Seems that I need some help. I tried to regenerate the files with `make build` like you suggested. Some packets were missing on my system (Arch Linux) so I install them. I noted that there is a `make dep` to build some dependencies. That command failed when building `node-sass`. Despite that `make build` worked after that. But I'm not happy with the result: https://gist.github.com/c-mauderer/0b0c84078bbcef9c85b9cd5857f54115

It seems that for some reason the generated code has different spaces. And it's quite a big block of code so it's not that easy to pick the parts that should have changed.

What is the recommended development host system (Ubuntu, CentOS, FreeBSD, Windows, ...?) and what host packages are necessary? Is there some documentation for that?

mpourismaiel:
The file looks good. If there isn't any build errors the code must be fine. We double check the demo to make sure everything is in order. We're adding automated tests to make the process a little easier. Thanks for your help.

I think Windows has a few problems with some of the features we're using so maybe avoid it while developing Syna. It might present some issues while building pages. Any Linux distro should be fine. I'm using Manjaro and it works great. We also only require yarn and hugo to be installed, all the other dependencies can be managed by running `yarn install` or `make dep`.

c-mauderer:
> The file looks good. If there isn't any build errors the code must be fine. We double check the demo to make sure everything is in order.

Thanks for your reply. I changed the commit message and pushed the patch on this branch.

PR #678, opened 2020-01-06
Partial fix to #677 #navbarCollapse

### LINKED ISSUE #677: [nav] regression in #navbarCollapse
bug

**What happened**:
Regression in nav

When you click on the nav hamburger it should show the list

Just click on the demo [site](https://about.okkur.org/syna/) with a small windows size

This is all I could find
### 0.15
**HTML:**
`<div class="navbar-collapse justify-content-between collapse show in" id="navbarCollapse"> `

**Function:**
assets/js/helpers/bootstrap-helper.js
```
function toggleMenu(node) {
  const menu = document.querySelector(node.dataset.target);
  menu.classList.toggle('in');
}
```

### 0.16
**HTML:**
`<div class="navbar-collapse justify-content-between collapse show" id="navbarCollapse">`

**Function:**
assets/js/collapse.js
```
const showCollapse = function (el, target) {
  $(el).attr('aria-expanded', 'true');
  $(el).removeClass('collapsed');
  $(target).addClass('show');
};
```

So the problem seems to be that in 0.16 the new functions don't add the **in** class

**Environment**:
- Syna Theme version:  0.16
- Hugo version:  0.60

Marzal:
Re-implement the class "in"

fixes #677 (partial)

Not sure if is the best/right way but this fix the hamburger menu regression.
It's not enough although.
`assets/scripts/syna-collapse.js` also needs updating, but I don't know if there is a proper method to do it better than by hand.

**Release note**:
```release-note
Fix #677  #navbarCollapse regression
```


Marzal:
Ok, after a bit of trial and error I think I got it.

I was missing the `node_modules` folder that a empty `yarn` downloads and then I could run
`yarn run build-js`

stp-ip:
There seems to have been a bigger underlaying issue with a code change leading to a full DOM reset resulting in lost event listeners. This should be fixed in #681.

Thanks for debugging and adding a fix, which unfortunately isn't necessary anymore with the bigger issue fixed. We'll do our best to add tests so such a regression won't happen again.

Closing for now and commenting on the issue itself.
Looking forward to more contributions especially ones we merge in ;)

PR #688, opened 2020-01-18
Fix list fragment not listing pages from all sections

### LINKED ISSUE #650: [blog] root path articles not possible
**What happened**:
Currently hosting articles on the root path doesn't work.

Setup articles within the first level.
blog/some-article vs /some-article.

This fails as far as I can see due to list only accepting a subpath such as "blog/" and not "/".
Also seems it does default to the current directory for _index, which is not what one wants as _index assumes all things are fragments not pages so each article will show up as a content file.

Needs a bit of thinking, if that is something to support. If this is a problem in our current thinking or what else might be doable.

mpourismaiel:
List fragment didn't list pages from some sections.

fixes #650 

**Release note**:
```release-note
- list: Fix list fragment not working on some sections
```

### DIFF
--- a/layouts/partials/helpers/list-pages.html
+++ b/layouts/partials/helpers/list-pages.html
@@ -1,6 +1,9 @@
 {{/* Sets page to either the given frontmatter section variable, current
   parent section or if the current page is a section, the current page */}}
-{{- .page_scratch.Set "page" (.Site.GetPage "Section" (.Params.section | default .root.Section)) -}}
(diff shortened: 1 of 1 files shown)


stp-ip:
Cypress is still failing on this.

mpourismaiel:
End to end tests take a bit of time to run in containers with limited resources and the time limit I've set for the test runs out. Rerunning the tests fixed the issue.

PR #696, opened 2020-01-25
Add support for icon in portfolio fragment

### LINKED ISSUE #649: [portfolio] Icon asset is not supported
bug

**What happened**:
The portfolio fragment does support the asset declaration, but not the icon from that.

Icon support for portfolio.

Set an icon instead of an image to asset.

Any other fragments that support asset, but not the full spec including icon?

mpourismaiel:
Add support for icons in portfolio fragment

fixes #649 

**Release note**:
```release-note
- portfolio: added support for icons
```


stp-ip:
On the modal we also need the icon support or at least remove the image code or it will load a broken image.

Besides that it looks good.

mpourismaiel:
It should be fixed. It's not really a good approach for the displaying of the icon in the modal but I can't think of anything else at the moment.

stp-ip:
`` Still shows up and breaks ;)

But the icon in the modal is already good. Happy with that solution, even if it's not perfect.

mpourismaiel:
My bad. Pushed a fix for that.

PR #697, opened 2020-01-25
wip: Add support for postcss with rtlcss and autoprefixer

### LINKED ISSUE #115: Add support for RTL stylesheet
feature

**What happened**:
Can not use another stylesheet instead of Syna's own stylesheet altogether. This causes users not being able to use RTL stylesheets or change the stylesheet from ground up.

Import another main stylesheet instead of Syna's main one. Removing the main stylesheet from the final build from the user side.

Try to use another stylesheet. I.E. run rtlcss on default stylesheet and use the resulting bundle.

**Environment**:
- Syna Theme version:  
- Hugo version:  
- Others:

mpourismaiel:
Add support for postcss, rtlcss and autoprefixer

fixes #115 

WIP

**Release note**:
```release-note

```


stp-ip:
Sometimes I love how tiny changes might make a huge impact.

A few things:
This needs outside software (postcss-cli) so making Syna harder to use and deploy even for the current use case.
My thinking would be at least for now. Keep ltr functioning as is and go the postCSS codepath in case of rtl. Does that seems reasonable/workable?

Keep existing way for ltr without postCSS keeping from firstly breaking current users as well as keeping the deployment and starting with Syna simple.

Additionally documentation the changes and deps necessary for rtl.

Some changes to netlify and cypress needed as well.

mpourismaiel:
We can create a config for it, removing it from the default and notify users who add the config that they need to install dependencies. Sadly Hugo doesn't integrate external build processes and features like this tend to become more complex.

stp-ip:
Closing this as we want to move this into documentation instead and PostCSS is enabled within the purgecss PR.

PR #699, opened 2020-01-30
Add eslint and prettier for js files

### LINKED ISSUE #698: Add linter and code style guide for JS files
Currently JS files have no style guides to follow and should be manually checked.

mpourismaiel:
Added linter and fixer for JS files

The configuration uses airbnb base config which is the closest to our style and prettier cleans up the code and doesn't interfere with airbnb's linting rules.

fixes #698


stp-ip:
Cypress failing ;)

mpourismaiel:
We have to add some dependencies for cypress. Seems like Github's container is missing some packages that need to be installed manually. I'll read about how we can do so, but I guess we need a container of our own for it. https://docs.cypress.io/guides/guides/continuous-integration.html#Advanced-setup

stp-ip:
Sounds fun ^^

PR #712, opened 2020-02-16
WIP: Add support for memberspace

### LINKED ISSUE #631: Add memberspace to Syna
feature

**What happened**:
Currently creating membership sites isn't possible with Syna.

Make membership sites possible with Syna.

Not really sure how this would work or even if it works.
Wanted to note it down.

mpourismaiel:
Add support for memberspace

fixes #631 

The implementation works with a fragment that you can add to _global or any page or section globals and in turn nav fragment in those pages will display a login link. I didn't add anymore customizations since I'm not even sure if this is the best approach. My assumptions is that the user would create a memberspace.md file in _global with following content:

```
+++
fragment = "memberspace"
subdomain = "subdomain_copied_from_installation_guide_of_memberspace"
+++
```

And then the nav fragment will display a link to Login/Signup page in Memberspace. It would be recommended that the button Memberspace provides be disabled from their website. But we can also move the "go to top" button as we discussed.

If you have any other approach in mind please let me know and if everything is okay I can add documentation and we can move forward with the merge.

**Release note**:
```release-note
- memberspace: New fragment that adds memberspace scripts to enable the service for Syna
```

### DIFF
--- a/layouts/partials/fragments/nav.html
+++ b/layouts/partials/fragments/nav.html
@@ -81,6 +81,13 @@
               </a>
             </li>
           {{- end -}}
+        {{- end -}}
+        {{- with ($page_scratch.Get "memberspace") }}
+          <li class="nav-item">
+            <a href="https://{{- .Params.subdomain -}}.memberspace.com/member/sign_in" class="nav-link anchor">
+              Login / Signup
+            </a>
+          </li>
         {{- end }}
       </ul>
       {{- if .Params.search }}
--- a/layouts/partials/helpers/fragments.html
+++ b/layouts/partials/helpers/fragments.html
@@ -184,7 +184,7 @@
         {{- if or $is_404 (ne .Params.fragment "404") -}}
           {{- if eq .Params.fragment "config" -}}
             {{- $page_scratch.Add "page_config" . -}}
-          {{- else if eq .Params.fragments "memberspace" -}}
+          {{- else if eq .Params.fragment "memberspace" -}}
             {{- $page_scratch.Set "memberspace" . -}}
           {{- else -}}
             {{- $page_scratch.Add "page_fragments" . -}}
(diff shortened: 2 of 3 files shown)


layouts/partials/helpers/fragments.html
```diff
@@ -172,6 +172,7 @@
   looking from the page directory or file to it's parents and so on. */}}
 {{- $page_scratch.Set "page_fragments" (slice) -}}
 {{- $page_scratch.Set "page_config" (slice) -}}
+{{- $page_scratch.Set "memberspace" nil -}}
```
stp-ip, line 175: Using a global variable for a vendor specific fragment seems problematic.

layouts/partials/fragments/nav.html
```diff
@@ -81,6 +81,13 @@
               </a>
             </li>
           {{- end -}}
+        {{- end -}}
+        {{- with ($page_scratch.Get "memberspace") }}
```
stp-ip, line 85: Adding a vendor specific fragment into another fragments code seems problematic.
We should inject it from a memberspace fragment or at least make this injection more generic.
As it currently stands this seems too bundled.

mpourismaiel:
Currently I can't think of any approach that would be less complex. I would love to hear ideas on the matter. Maybe remove the fragment and make it a site configuration, applied to the baseof template?

JOduMonT:
I'm curious, it is like a paywall ?
What manage the authentication Discourse, Netlify ?

stp-ip:
Memberspace is a paywall service especially for membership use cases. They have different modes of security. Simplest will remove content until payment via JS, one will add the content in via JS from their DB after payment/login.

Authentication is therefore done on the memberspace side of things.

JOduMonT:
I didnt knew it was a SaaS
I wonder if these bypass works ?
- https://github.com/iamadamdev/bypass-paywalls-chrome
- https://github.com/iamadamdev/bypass-paywalls-firefox

stp-ip:
Most likely depends on the mode used.
But that's not part of Syna so not really relevant here.

stp-ip:
Let's shelf this for now. We first need a better idea around provider fragments and also can't add things globally for a fragment.

Possible option for injecting something like this is the fragments JS or our event framework. Let's discuss this separately.

PR #738, opened 2020-04-12
Fix date_format config

### LINKED ISSUE #736: date_format param does not work for publish dates although used in partial
bug

**What happened**:
I set following in `config.toml`:

```
[params]
  date_format = "02.01.2006"
```

But this did not change the date format.

Since the `publish-date.html` helper partial contains this
```
{{- $date.Format ($.site.Params.date_format | default "2006-01-02") -}}
```
I expected the date format to change.

Just do the same config as I did

If I understood it right, partials seem to have an isolated scope and no access to outer scope. This could mean it's not possible at all inside a partial to use `$.site.Params`. This could be the root-cause.
See https://gohugo.io/templates/partials/#variable-scoping

**Environment**:
- Syna Theme version: `v0.16.2`
- Hugo version: `Hugo Static Site Generator v0.68.3/extended darwin/amd64 BuildDate: unknown`
- Others:

stp-ip:
Fixes date format config not being used. $.Site wasn't the correct entrypoint, but should have been .root.Site due to it being injected into the partial from us.

fixes #736

```release-note
Fix global date_format config not being applied
```

### DIFF
--- a/layouts/partials/helpers/publish-date.html
+++ b/layouts/partials/helpers/publish-date.html
@@ -1,4 +1,6 @@
 {{- $date := .root.Date | default .root.Lastmod }}
+{{- $date_format := .root.Site.Params.date_format | default "2006-01-02" }}
+
 {{- if not (eq .badge false) }}
   <span class="publish-date badge mr-1
     {{- if (eq $.background "dark") -}}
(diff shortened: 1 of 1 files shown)


stp-ip:
It's due to the partial scope not having access to the global `$` scope. That was the reason we introduced $self and $root, which we long term should probably make consistent ^^

PR #755, opened 2020-05-04
Fix duplicate section pages in list fragment

### LINKED ISSUE #751: Investigate duplicate pages within list
bug

**What happened**:
With the default `subsection = true` list will show duplicated subsections.

Single subsections.

Create a list and content like this:
```
draft/
_index.md
  _index/list.md
  _index/index.md
  posts/
    _index.md
    test-post-1.md
    test-post-2.md
```
It will show "posts" twice in the list.

**Environment**:
- Syna Theme version:  v0.17.0

mpourismaiel:
`.Pages` previously only returned pages with `.Kind eq "page"` but now it returns all pages under a section. Changed our list helper to collect pages based on their `.Kind` if `subsections` is not set or is set to false.

fixes #751

**Release note**:
```release-note
- list: Fix a bug that caused subsections be shown if the config was set to not show them or show duplicate subsections if config was set to show them.
```

### DIFF
--- a/layouts/partials/helpers/list-pages.html
+++ b/layouts/partials/helpers/list-pages.html
@@ -10,9 +10,11 @@
   {{- end -}}
 {{- end -}}
 
-{{- .page_scratch.Set "pages" (.page_scratch.Get "page").Pages -}}
-{{- if .Params.subsections -}}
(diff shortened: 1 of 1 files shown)


mpourismaiel:
A good place to check this out: `/dev/list/`.

PR #772, opened 2020-05-23
Spanish translation

Marzal:
Add Spanish translation and one missing item

**Release note**:

```release-note
Added Spanish translation
```


stp-ip:
Awesome. Good catch on the english ones being missing.

PR #774, opened 2020-05-27
Fix broken if and statement

### LINKED ISSUE #773: Theme fails to build in hugo version 0.71.1 (0.71.0 works fine)
bug

**What happened**:
Attempting to build site with syna theme with hugo v0.71.1 results in error message:

```
Error: Error building site: failed to render pages: render of "page" failed: "/path/to/project/themes/syna/layouts/_default/baseof.html:10:9": execute of template failed: template: _default/single.html:10:7: executing "_default/single.html" at <partial "helpers/fragments-renderer.html" (dict "page_scratch" .Scratch)>: error calling partial: "/path/to/project/themes/syna/layouts/partials/helpers/fragments-renderer.html:10:9": execute of template failed: template: partials/helpers/fragments-renderer.html:10:9: executing "partials/helpers/fragments-renderer.html" at <(not (isset .Params "slot")) (ne .Params.slot "")>: can't give argument to non-function not (isset .Params "slot")
```

Site builds correctly

* Install theme on site
* Try and build site with hugo v0.71.1

Works fine with hugo v0.71.0.  Unsure if this is syna doing something it shouldn't or this is a regression on hugo's side.  Scanning through the code in the latest changelog, and looking at the error message, it seems like it's related to [these changes](https://github.com/gohugoio/hugo/commit/6c3c6686f5d3c7155e2d455b07ac8ab70f42cb88#diff-981fbfeb374ac30e53aac700cbc33234L32). (that link may not link correctly due to github not wanting to render the diff by default, look for file " tpl/internal/go_templates/texttemplate/funcs.go")

Probably going to open an issue with hugo as well given that it seems like a regression on their side.

**Environment**:
- Syna Theme version:  
sha 5bbd985
- Hugo version:  
0.71.1
- Others:

Houndie:
Updates to hugo 0.71.1 causes a previously allowed-but-wrong construct to now be disallowed.  This fixes the template to now compile on 0.71.1

Fixes okkur/syna#773

Fixes an template that is now illegal in hugo 0.71.1

Also discussed in gohugoio/hugo#7325
**Release note**:

```release-note
Fixes template error in hugo 0.71.1
```

### DIFF
--- a/layouts/partials/helpers/fragments-renderer.html
+++ b/layouts/partials/helpers/fragments-renderer.html
@@ -7,7 +7,7 @@
 {{- range sort ($page_scratch.Get "page_fragments" | default slice) "Params.weight" -}}
   {{/* If a fragment contains a slot variable in it's frontmatter it should not
     be rendered on the page. It would be later handled by the slot helper. */}}
-  {{- if (not (isset .Params "slot")) (ne .Params.slot "") -}}
+  {{- if and (not (isset .Params "slot")) (ne .Params.slot "") -}}
     {{/* Cleanup .Name to be more useful within fragments */}}
     {{- $name := cond (eq $page .) .File.BaseFileName (strings.TrimSuffix ".md" (replace .Name "/index.md" "")) -}}
     {{- $bg := .Params.background | default "light" }}


stp-ip:
Thanks for the fix. Looking forward to the next PR ;)

stp-ip:
Referencing the upstream Hugo issue here as well as the fix was merged before the guidance on PR linking: https://github.com/gohugoio/hugoThemes/issues/858

PR #782, opened 2020-06-16
Fix nav fragment to allow multiple navs per page

### LINKED ISSUE #388: [nav] Search doesn't work properly if there are more than one navbar on the page
bug

**What happened**:
Search in navbar doesn't work properly

To be able to search using any navbar available on the page

Add more than one navbar to the page with search activated on them

**Environment**:
- Syna Theme version:  v0.13
- Hugo version: v0.48
- Others:

mpourismaiel:
Fixed the search script in the nav fragment, allowing multiple nav fragments to use search in the same page.

fixes #388 

The sticky option can be used but we may need to rethink how it might function differently for multiple nav fragments. Right now if you have two sticky navs, they are followed by each other on top of the page. If we need a different behavior, I suggest doing so in a separate issue.

**Release note**:
```release-note
- nav: Fix the bug that prevented search from working if there were multiple nav fragments in one page
```

### DIFF
--- a/cypress/integration/navbar.spec.js
+++ b/cypress/integration/navbar.spec.js
@@ -27,4 +27,19 @@ describe("Navbar", () => {
     cy.get('[data-toggle="collapse"][data-target="#navbarCollapse"]').click();
     cy.get("#navbarCollapse").should("be.hidden");
   });
+
+  it("should show search results with multiple navs in page", () => {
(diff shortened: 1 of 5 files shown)


stp-ip:
We should have a dev page testing this as well as a test that it works. Overall code looks good.

mpourismaiel:
Created dev page and test accordingly.

PR #784, opened 2020-06-16
Slice list fragment

### LINKED ISSUE #783: Refactor fragments to make development easier
**Refactor request**

Currently fragments have a lot of functionality spread between their code that makes reading their code hard. Also some fragments have different variants which adds to the problem. We have helpers to reduce this problem but it's not helping by that much.

My proposal is to slice up fragments into smaller parts that don't change between variants and also separate different variants of each fragment.

This will allow us to change a part of the fragment much easier and faster and only care about how it impacts the fragment visually. Separating variants will also make the code for each fragment much smaller.

A nice side effect to this is that users can override a fragment and any slice that they haven't changed will still get updates, as opposed to before that a change to a fragment would mean opting out of all future updates to that fragment.

mpourismaiel:
Slice list fragment's code and separate it's different variants.

fixes #783 

This is in line with #783 proposal. Please test the demo extensively since the code has changed quite a lot and even though I have tried to test it's every aspect, something might have skipped my eye.

**Release note**:
```release-note
- Simplify fragment code base by separating different variants and modularizing them (list)
```


stp-ip:
On a first look this seems like a reasonable approach, but I want to take another look later tomorrow and iterate especially on the ideas mapping to hugo modules.

stp-ip:
As discussed. I think this additional layer of abstraction adds too much complexity in comparison to the reduction of code duplication.
Different design framework might do different things and I think using helpers and per framework partials is the better way to go.

Awesome thought and happy you are coming up with new solutions and ideas.

Closing this in favor of duplication and a sub directory for each framework and a specific fragment call such as `fragment = "tailwind/contact"`.

Let me, know if that doesn't match, what we discussed.

Issue #513, opened 2019-01-25
[nav] With sticky enabled content is not perfectly visible


stp-ip:
**Is this a BUG REPORT or FEATURE REQUEST?**:
bug

**What happened**:
Sticky = true for nav makes the text unreadable: https://syna-demo.okkur.org/docs/global-variables/#fragment

**What you expected to happen**:
Scroll the height of the nav more, if sticky is set.

**How to reproduce it (as minimally and precisely as possible)**:
https://syna-demo.okkur.org/docs/global-variables/#fragment

mpourismaiel:
This is a really hard problem to fix. I can't think of any working solutions. There is a solution that only work with in-page-transitions but that's not really good since most of the time the transition would be between pages.
 
Checkout react's doc that has the same problem. [reactjs.org/docs/optimizing-performance.html#brunch](https://reactjs.org/docs/optimizing-performance.html#brunch)
MDN has sort of the same problem, they have fixed it with `margin-top`. Can work but will break with custom styling: [developer.mozilla.org/en-US/docs/Web/API/Document/createElement#Examples](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement#Examples)
Kubernetes has removed the navbar in their doc page where in page transition is active

PR #788, opened 2020-06-24
Fix sticky sidebar issues

### LINKED ISSUE #601: [content]: Sidebar is not scrollable
bug

**What happened**:
When sidebar is sticky and there is ToC fragment in it, it's not scrollable.

The sidebar should be scrollable even if it's sticky.

Enable sticky option in sidebar config and use ToC inside it.

**Environment**:
- Syna Theme version:  `0.15.0`
- Hugo version:  `Hugo Static Site Generator v0.55.0-4333CC77F/extended linux/amd64 BuildDate: 2019-04-08T17:05:23Z`
- Others:

### LINKED ISSUE #700: Sidebar with sticky = true
bug

**What happened**:
The sidebar sticky = true doesn't stick !

> content.html (**content** fragment)

if you choose `sticky = true` 
for the sidebar in a content fragment. The sidebar doesn't stick.

the side should stick when one scroll down.
but it doesn't !

To reproduce get a content fragment with a sidebar (either left or right)
set the sticky Boolean to true, check in the source code that the class 

> sticky-top pt-3

is set for the child element of `content-sidebar`
Then scroll down, you'll see that the element is not sticky.

I tried with 

- Brave Version 1.2.43 Chromium: 79.0.3945.130 (Official Build) unknown (64-bit)
- and Firefox 72.0.2 (64-bit). 

I think that the problem comes from the fact that the parent element of #sticky-top  div :
aka : 
```
.content-sidebar {
  word-break: break-word;
  overflow-y: auto;
}
```
has got the style :   **overflow-y: auto;**

I made it work by adding one div around the first _sticky-top_ :

```
        {{- if .sticky }}
          <div class="sticky-top pt-3"> <div class="pt-3">
        {{- end -}}
```
[...]

```
    {{- if .sticky }}
          </div></div> {{/* .sticky-top */}}
          <div></div>
        {{- end -}}
```
and I add in _syna.css:

```
.sticky-top{
    top: 50px;
}
```
and remove **overflow-y** from :

```
.content-sidebar {
  word-break: break-word;
/*   overflow-y: auto; */
}
```

**Environment**:
- Syna Theme version:  v0.17.0
- Hugo version:  Hugo Static Site Generator v0.63.2/extended linux/amd64 BuildDate: unknown
- Others: Arch linux

mpourismaiel:
- Fix sticky sidebar scroll issue (Content was not scrollable when it exceeded height of the page)
- Fix sticky sidebar being overlapped by sticky navbar

fixes #700 
fixes #601 

**Release note**:
```release-note
- content: Fix sticky sidebar issues
```


mpourismaiel:
I forgot to mention, please scroll the docs page with some of the articles expanded, in a long page with browser height set to 400px for example. The sidebar should stay in place while you scroll until it meets the header.

The code is implemented in a way that the two fragments stay independent and ignorant of each other. The content fragment only applies the scroll code if there is an active sidebar and should not break if there is a navbar or not.

mpourismaiel:
Fixed the bug that was causing the failing test and then fixed a faulty test that failed occasionally. It was failing for this PR so pushed the commit here, let me know if you want me to separate it.

layouts/partials/fragments/nav.html
```diff
@@ -224,3 +224,10 @@
   });
 </script>
 {{- end -}}
+{{- if .Params.sticky }}
+<style>
+  :root {
+    --header-height: 56px;
+  }
+</style>
+{{- end -}}
```
stp-ip, line 233: Not sure I like the hardcoding of that. Can't think of a better solution so.

layouts/partials/fragments/content.html
```diff
@@ -62,3 +62,28 @@ <h3 class="sidebar-title {{- printf " text-%s" $title_align -}}">
...
+      var extraPadding = 0;
+
+      if (sidebarTop <= headerHeight) {
+        extraPadding = headerHeight;
+      }
+
+      sidebarContentParent.style.setProperty('padding-top', extraPadding + sidebarContentParentPadding + 'px', 'important');
+    });
+  })();
+</script>
```
stp-ip, line 89: Why inline JS instead of within our JS scripts?

stp-ip:
Like the scrolling, but the position doesn't seem perfect. This is more a side effect of #513 I guess. As we are overwriting the height. Would that be something we would like to fix here as well or am I missing something?

stp-ip:
Edgecase, that should probably get a dev/ page and maybe a test. What happens with 2 sidebars? Do we need an additional note for the "sticky" config for the content fragment?

layouts/partials/fragments/nav.html
mpourismaiel: I can't think of a solution that doesn't involve JS and is not hard coding the value. I'll try to figure something out and maybe we can create an issue to track it.

stp-ip: Fine to keep it as is and open a "revisit" to track this.

stp-ip:
What do you think about a dev/ page for manual testing and probably also automated testing?

mpourismaiel:
@stp-ip Added tests and dev pages accordingly.

PR #800, opened 2020-07-15
Update subitems helper to fix multilingual issue

### LINKED ISSUE #799: Multilingual mode and page of type : items / faq
bug ? (I dont know if is a bug or if I do something wrong)

**What happened**:
In the list fragment I have : 

 - services/index.fr.md
 - services/index.fr.md
 - services/list-tiled.md
 - services/list-tiled.fr.md

In the list-tiled.* I have defined the source : 
section = "services-folder1"

In the services-folder1 I have : 
 - sample1.md
 - sample1.fr.md
 - sample2.md
 - sample2.fr.md

It work fine in french and english

If I try to do the same with the faq or items fragments, that dont work, the page is simply not rendered and in place the home page is shown.

**Environment**:
Syna Theme version: v0.17.2
Hugo version: 0.72

Thanks Cédric

mpourismaiel:
`subitems` helper didn't identify page resources correctly in multilingual websites. The problem appeared in not default language pages.

fixes #799 

**Release note**:
```release-note
- Fix multilingual pages not showing subitems of fragments in non default language page
```

### DIFF
--- a/layouts/partials/helpers/subitems.html
+++ b/layouts/partials/helpers/subitems.html
@@ -3,7 +3,9 @@
 {{- $name := .Name -}}
 {{- $site := .Site -}}
 {{- $page_scratch := .page_scratch -}}
-{{- $items := where ($page.Resources.Match (printf "%s/*.md" $name)) ".Name" "ne" $self.Name -}}
+{{ $default_lang := .root.Site.Params.DefaultContentLanguage | default .root.Site.Language.Lang }}
+
+{{- $items := where ($page.Resources.Match (printf "%s/*.md" (strings.TrimSuffix (cond (ne $default_lang $page.Language.Lang) (printf "/index.%s" $page.Language.Lang) "") $name))) ".Name" "ne" $self.Name -}}
 
 {{- if eq (len $items) 0 -}}
   {{- range $directory := ($page_scratch.Get "fragment_directories") -}}


stp-ip:
So what various ways of multilingual config work here?
`faq.<lang>.md + faq/column1.<lang>.md`
but not
`faq.<lang>.md + faq.<lang>/column1.md`

Do we wanna be specific with each column having the language or the subdirectory? What do you think?
Also needs better docs on how to do this or?

mpourismaiel:
`faq/index.<lang>.md + faq/column1.<lang>.md` and `faq.<lang>.md + faq.<lang>/column1.<lang>.md` work. The best description I can come up with is that Hugo uses the language specific files in non default languages and assumes all other files and pages to be specific to the default language. In other words, user should create each page and resource for each language if they want that file or page to be available in that language. You can have pages for a language but not for the other and you can have pages for both languages, and the naming convention assumes pages without `.<lang>` to belong to the default language.

Definitely requires more documentation since I seem to forget this everytime I create multilingual websites.

Language directories are used for this very purpose but implementing them and not breaking most of our code is a challenge. I'll approach it some other time but can't think of a solution for it now.

stp-ip:
Ok let's create an issue to get this documented.

Thanks for reminding me that the language directories is still on our radar, but not implemented. 

lgtm

PR #803, opened 2020-07-18
Add new repositories fragment

### LINKED ISSUE #599: [cards-repo] Add repo cards for items
feature

**What happened**:
We currently use items/item/portfolio for showcasing projects etc.

Have an additional card, which can be used with externally propagated data.

Interesting repository list example. http://sap.github.io

mpourismaiel:
Added new fragment called "repositories". Demo and documentation is available at `/fragments/repositories`. Currently it only supports Github. I plan to add Bitbucket and Gitlab shortly but adding them would require a simple interface to keep us from writing the template multiple times. Writing the code isn't that hard but I prefer to get approval on the design and how it's currently implemented before I move on.

image

fixes #599 

**Release note**:
```release-note
- repositories: Added new fragment to fetch and display user repositories from Github
```


assets/styles/_syna.scss
```diff
@@ -227,3 +227,15 @@ section .bg {
...
+  .repository {
+    display: flex;
+    justify-content: stretch;
+    align-items: stretch;
+
+    .card {
+      flex: 1;
+    }
+  }
+}
```
stp-ip, line 241: Why add the repositories styles to the global syna config?

exampleSite/content/fragments/repositories/docs.md
```diff
@@ -0,0 +1,32 @@
...
+*type: number*
+*default: 1000*
+
+The number of repositories that will be rendered.
+
+#### sort
+*type: number*
+*default: updated_at*
+
+The variable name to use for sorting repositories.
```
stp-ip, line 28: Why "variable name"?

exampleSite/content/fragments/repositories/docs.md
```diff
@@ -0,0 +1,32 @@
...
+The username of the user you want the repositories of.
+
+#### count
+*type: number*
+*default: 1000*
+
+The number of repositories that will be rendered.
+
+#### sort
+*type: number*
```
stp-ip, line 25: Why is this a number?

exampleSite/content/fragments/repositories/docs.md
```diff
@@ -0,0 +1,32 @@
...
+*default: updated_at*
```
stp-ip, line 26: What options are available? Star count? Contributors?

exampleSite/content/fragments/repositories/index.md
```diff
@@ -0,0 +1,18 @@
...
++++
+
+Display user repositories from Github.
+
+<!--more-->
+
+<div class="alert alert-primary" role="alert">
+{{% md %}}
+For now, we only support Github. You can review the [documentation for their API](https://docs.github.com/en/rest/reference/repos#list-repositories-for-a-user).  
+We will add support for more sevices in the future.
```
stp-ip, line 16: Not with the current configuration adding more services would be possible. It feels quite focussed on github. Adding other services would need more data from the user such as sub items with link to project on say gitlab or github.

We could leave it this simple and make it a first provider fragment: "repositories_github" for example.

exampleSite/content/fragments/repositories/repositories.md
```diff
@@ -0,0 +1,7 @@
++++
+fragment = "repositories"
+weight = 100
+user = "octocat"
```
stp-ip, line 4: We should probably showcase "okkur" for good measure :)

layouts/partials/fragments/repositories.html
```diff
@@ -0,0 +1,22 @@
...
+{{ $repositories := getJSON (printf "https://api.github.com/users/%s/repos" .Params.user) }}
+
+{{- partial "helpers/container.html" (dict "start" true "in_slot" .in_slot "bg" $bg "Name" .Name "Params" .Params "section_class" "content-fragment" "container_class" "overlay") -}}
+  <div class="row no-gutters repositories">
+    {{ range (first (.Params.count | default 1000) (sort ($repositories | default slice) (.Params.sort | default "updated_at") "desc")) }}
+      <div class="col-md-6 col-12 p-2 repository">
+        <div class="card">
+          <div class="card-body">
+            <h5 class="card-title">{{ .name }}</h5>
+            <h6 class="card-subtitle mb-3 text-muted">{{ printf "Last updated on: %s" (dateFormat "02/01/2006" .updated_at) }}</h6>
```
stp-ip, line 12: Why not use the ISO date format aka "2000-12-31"?

mpourismaiel: I somehow forgot to use the standard format :D Will update it.

exampleSite/content/fragments/repositories/index.md
mpourismaiel: So rename the fragment to repositories_github and remove the alert?

I was thinking about adding a `provider` and `framework` (or `variation`) variables instead of naming convention. It's not related to this issue though, just wanted to mention it somewhere since I tend to forget a lot :D

exampleSite/content/fragments/repositories/docs.md
mpourismaiel: I can't think of a good explanation, not happy with it either. Any suggestions?

exampleSite/content/fragments/repositories/docs.md
mpourismaiel: Totally my bad.

mpourismaiel:
I don't think subitems would be a good idea for this fragment. A whitelist of repositories can be created using a variable, specifying fields is something that is applied on all the cards (I don't think changing the fields per project is a good idea, both design-vise and the confusion that it causes to the end user). Also we can allow a filter function to be written by the user and call that. For example, we write our sort and filter functionality in a helper `partials/helpers/repositories_github/fitler.html` that returns a filtered list and document how to write new filters. We can keep the sort and count in our code for the sake of simplicity as well.

stp-ip:
Ok so we keep it simple.
Make it the first provider specific fragment.
Adding more buttons from the data as follow on.

So default sorting is just by last commit date.
We should have at least star count as well. Sidequestion there. Does a fork have the same star count as the main repo? If not, then filtering wouldn't be needed. We would just have star count default sort and it would always bubble up the users personal repos and show the 5 most important or so.

Not sure about what you are imagining on the filter side. Care to elaborate? My first thought was just whitelisting of repos as a list.

Thinking about lists. Any reason to not make the "user" not a list? Also is this better to be called "user" or namespace?
It being a list of namespaces for example we could do something like: "okkur, okkur-incubator".

The other option there is to keep it simple and pull in only one namespace/user and the Syna user could always create multiple separate fragments.

mpourismaiel:
Sorting and filtering is a bit of a problem since we can't get pinned repositories and have to rely on sorting and limiting or filtering (based on a whitelist or "no forks" somehow) and again, limiting with count. We can however extract the code we use to filter repositories into a helper file that receives repos as a parameter and returns the new list. Then any user who wishes to change the list to their liking can override that helper, copy it to their website and update its code.

stp-ip:
I would leave advanced sorting and filtering for now, but add these points:
* sorting by date (what we have)
* sorting by star count (new default)
* forks true/false (false by default)
* whitelist that only shows repos that match the whitelist (not set by default)

Happy to do the whitelist idea in a separate PR. The others I feel are good things to have here already. Thoughts?

mpourismaiel:
image

exampleSite/content/fragments/repositories_github/docs.md
```diff
@@ -17,16 +17,22 @@ The username of the user you want the repositories of.
...
-*type: number*
-*default: updated_at*
+*type: string*
+*default: stargazers_count*
 
-The variable name to use for sorting repositories. 
+This key is used to sort the repositories.
 
-**Note:** The variable name is a key in the response of Github's repositories API. Also the sort is done using Hugo's `sort` function.
+**Note:** This value is a key from the response of Github's repositories API. Also the sort is done using Hugo's `sort` function.
```
stp-ip, line 36: So can be any key from the Github response. Nice.

stp-ip:
Let's do the whitelist and maybe additional buttons in a separate PR.

PR #816, opened 2020-08-01
Fix postcss issues

### LINKED ISSUE #813: Error building site with v0.17.3
bug

**What happened**:

With update of my webpage to Syna v0.17.3, the following error started to occur while running `hugo server`:

```
Building sites … WARN 2020/07/26 19:34:36 .File.Dir on zero object. Wrap it in if or with: {{ with     .File }}{{ .Dir }}{{ end }}
Built in 449 ms
Error: Error building site: POSTCSS: failed to transform "style.css" (text/css): resource "scss/styles/index.scss_02bf14281ac065540fcc7ecc15ca70a4" not found in file cache
```
BTW the same happens also on your [okkur / syna-start](https://github.com/okkur/syna-start) project.

No error occurs.

1. Checkout syna-start with
```
git clone --recurse-submodules https://git.okkur.org/syna-start
```
2. Run `hugo server`.

**Environment**:
- Syna Theme version:  v0.17.3
- Hugo version: Hugo Static Site Generator v0.74.3/extended windows/amd64 BuildDate: unknown
- Others: OS Windows 10 Pro 64-bit

mpourismaiel:
Disable postcss asset pipes by default, it requires `postcss` variable to be set to `true` in site's `Params` config.

fixes #813 

**Release note**:
```release-note
- Fix issue preventing site build when `postcss-cli` is not installed
```

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -52,8 +52,23 @@
   {{- $sassTemplate := resources.Get "styles/index.scss" -}}
   {{- $options := (dict "targetPath" "style.css" "outputStyle" "compressed" "enableSourceMap" true "includePaths" (slice "node_modules" "styles")) -}}
   {{- $styleTemplated := $sassTemplate | resources.ExecuteAsTemplate "main.scss" . -}}
-  {{- $style := $styleTemplated | resources.ToCSS $options | resources.PostCSS | resources.Minify | resources.Fingerprint | resources.PostProcess }}
-  <link href="{{ $style.Permalink }}" rel="stylesheet">
+  {{- if hugo.IsProduction -}}
+    {{- if .Site.Params.postcss -}}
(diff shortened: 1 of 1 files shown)


stp-ip:
Rerunning the test.

stp-ip:
Interesting. We are running into the issue we want to fix in the tests. Did you run the tests locally as well?

mpourismaiel:
I'm not really sure what's happening. Hugo builds and runs development server in all conditions. Production and development both work without `postcss = true` and they fail if `postcss = true` is set and postcss is not installed. But they will run with that parameter set, if postcss is installed. 

In the test environment, cypress is not running since Hugo is failing and I can't figure out why that's happening. I'd appreciate any ideas or solutions.

stp-ip:
Haven't figured out what's wrong here. Will take another stab at it once time permits.

nzt-ventures (approved):

PR #818, opened 2020-08-06
Update dependencies

mpourismaiel:
**What this PR does**:
Update dependencies and update build files. Checked pages to make sure everything looks good and works fine but please, check them again. Something might have broken and slipped through tests and my examination.

**WIP**:

- [x] exampleSite packages to be updated
- [x] Check if any package is unused or not needed
- [x] Check if anything is broken

Also, we need more tests. With every new feature/refactor/bug PR my blood pressure goes to roof, fearing I might have broken something and not realized it. :D

**Release note**:
```release-note
- Updated JS dependencies
```


stp-ip:
Definitely happy with more tests :+1:

PR #832, opened 2020-10-13
fix issue 831 : read more link on list page stays on the same page

### LINKED ISSUE #831: Read more link stays on the same page
Hi,

Trying to make a blog with Syna, I found that "Read more" link after the summary of an article sends to the current page (articles list) instead of the article page.
Clicking on the title works correctly, displaying the full article.

You can see the bug on okkur.org blog : https://blog.okkur.org/articles/ 

Thank you for this nice theme.
Damien.

damienjegou:
fixes #831

### DIFF
--- a/layouts/partials/fragments/list.html
+++ b/layouts/partials/fragments/list.html
@@ -128,7 +128,7 @@
                           {{- else -}}
                             {{- printf " badge-%s" "secondary" -}}
                           {{- end -}}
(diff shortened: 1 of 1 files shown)


stp-ip (approved):
Great. Thanks for the PR and looking forward to the next one :P

Sidenote: Might take a bit to propagate through to the (okkur) blog and docs pages you used for showcasing the bug.

PR #854, opened 2021-06-07
Add missing translation strings

Gilrain:
Doing a 'grep -R i18n layouts' revealed some unlisted elements.


stp-ip (approved):
Thanks. That seems to have been missed.

/lgtm

PR #856, opened 2021-06-07
Display correct error on contact page

Gilrain:
When a user doesn't write a message on the contact page, the wrong error is displayed: 'Send Message' instead of 'Please enter a message'.


stp-ip (approved):
Good catch thanks
/lgtm

PR #863, opened 2022-02-01
New social buttons on navbar (#862)

### LINKED ISSUE #862: New social buttons on navbar

feature

**What happened**:

It would be nice to have nor social buttons on the navbar such as LinkedIn, Facebook, Twitter and also a custom option.

If the new buttons are declared in the nav definition, they should appear beside the **repo_button**. For example:

```
[repo_button]
  url = "https://github.com/okkur/syna"
  text = "" # default: "Star"
  icon = "fab fa-github" # defaults: "fab fa-github"
  no_text = false # If true, the text will not be displayed

[linkedin_button]  
  url = "https://www.linkedin.com/"
  text = "" # default: "Star"
  icon = "" # defaults: "fab fa-linkedin"
  no_text = true # If true, the text will not be displayed
```
image

It also enhance the buttons by adding the parameter **no_text** that disables the text and displays a button with no text, just the icon properly centered inside the button. This is interesting because the current implementation does allow the use if the icon without a text by setting the text to " ". This, however, put the icon aligned to the left of the button instead of centering it.

N/A.

We already implemented this feature in a branch and we plan to submit it to the main repository as a pull request. Example:

```
[repo_button]
  url = "https://github.com/okkur/syna"
  text = "" # default: "Star"
  icon = "fab fa-github" # defaults: "fab fa-github"
  no_text = false # If true, the text will not be displayed

[linkedin_button]  
  url = "https://www.linkedin.com/"
  text = "" # default: "Star"
  icon = "" # defaults: "fab fa-linkedin"
  no_text = true # If true, the text will not be displayed

[twitter_button]
  url = "https://www.twitter.com/"
  text = "" # default: "Star"
  icon = "" # defaults: "fab fa-linkedin"  
  no_text = false # If true, the text will not be displayed

[facebook_button]
  url = "https://www.facebook.com/"
  text = "" # default: "Star"
  icon = "" # defaults: "fab fa-linkedin"    
  no_text = false # If true, the text will not be displayed

[custom_button]  
  url = "https://www.custom.com/"
  text = "" # default: "Star"
  icon = "" # defaults: "fas fa-star"  
  no_text = false # If true, the text will not be displayed
```
This results in:

image

**Environment**:

- Syna Theme version:  Latest version in the repository with the commit d086af4.
- Hugo version:  Tested with 0.80.0 and 0.76.5

fjtc:

Implements the features describe in the issue #862 by adding more social buttons in the nav bar.

We changed the example nav.md and the fragment nav.md to implement this issue. As a complement, I updated the English, French and Portuguese translations.

**Release note**:

```release-note
- Adding new social buttons to the nav bar (LinkedIn, Twitter, Facebook and custom) to be used just like the repo_button;
- The social buttons inside the nav bar now define a new parameter called no_text that removes the text label and centers the icons inside the button;
```


i18n/en.yaml
```diff
@@ -7,6 +7,22 @@
...
+- id: twitter.stars
+  translation: "Stars"
+- id: facebook.star
+  translation: "Star"
+- id: facebook.stars
+  translation: "Stars"
+- id: custom.star
+  translation: "Star"
+- id: custom.stars
+  translation: "Stars"
```
stp-ip, line 25: Yeah button text should probably not be using the key "star" as this only makes sense for Github starring. Maybe a more neutral `id: facebook.button_text` could work as well as the translation being being `Facebook`.

stp-ip: Same for the other languages.

layouts/partials/fragments/nav.html
```diff
@@ -129,11 +129,53 @@
...
+      </a>
+    </div>
+  {{- end }}
+  {{- with .Params.linkedin_button }}
+    <div class="d-none d-lg-block pl-auto">
+      <a class="btn btn-light ignore-color-change" href="{{ .url }}" role="button" title="{{- i18n "navbar.linkedin" | default "linkedin" -}}">
+        <i class="{{ printf "%s" .icon | default (printf "fab fa-linkedin") }} {{ if not .no_text }}mr-2 {{ end }}"></i>
+        {{- if not .no_text -}}
+          {{ .text | default (i18n "linkedin.star") }}
+        {{- end -}}
```
stp-ip, line 165: This is a bit manual and could be structured as a list. Any ideas on making it a list config wise as well as loop through the list in the code while keeping the backwards compatibility with the repo_button?

stp-ip (changes requested):
Happy to see this PR.
Added a few thoughts and open to consider such a change with a few edits.
One issue design wise is the non existent space between the buttons. It needs a bit of space between the buttons. Could be added with something like `ml-1` or so.

layouts/partials/fragments/nav.html
fjtc: I'm really sorry for my lack of experience with Hugo templates. I'm quite new to Hugo templates. I actually started to mess with it yesterday using Syna as a way to better understand Hugo. 

As you said, this patch is really to cumbersome to be useful so I'm planning to change it a little bit. In fact, today I notice that there is a better way to extend the template to make those icons more generic and, at same time keeping the backward compatibility with the existing **repo_button**.

My new plan is to left **repo_button** as is and add a new list of **button** that will add the new buttons in the same way **repo_button** works. This way we will be able to combine both **repo_button** and the new  **button** options to keep backward compatibility or delegate the buttons to use only the new **button** instead. I'll also try to make shortcuts for the most common icons to make the usage easier.

I also notice that there are more places in the sample site that I must update in order to document the new features and I'm planning to update those locations as well.

Could you reject this pull request for now? As soon as I have a better solution I will send a new one. We are really glad to be able to contribute to Syna's main repository.

Thanks in advance!

stp-ip: Sure no worries. Will close this one and wait for the next. Happy to help if anything comes up.
In terms of buttons that iterate take a look at this part of Syna: https://github.com/okkur/syna/blob/master/layouts/partials/fragments/buttons.html
Those are the call to action buttons, but it does showcase a more generic way of how it could work with Hugo.

i18n/en.yaml
fjtc: OK. I'll incorporate those changes as well.

PR #866, opened 2022-02-11
Adding pricing red icons css with example.

fjtc:

Adding  the ability to make  the pricing icons red.

image

This one was achieved by adding a new style to `_pricing.css` called `red`.  I added an example about how to use it inside the exampleSite's `plan-1.md`. I'm not sure if it is the preferred way to achieve this effect but I'm pretty sure it is the less disruptive one.

**Release note**:

```release-note
- Adding the ability to use red icons inside the pricing features;
```

### DIFF
--- a/assets/styles/_pricing.scss
+++ b/assets/styles/_pricing.scss
@@ -39,6 +39,10 @@
       flex: 1 0 auto;
     }
 
+    .red {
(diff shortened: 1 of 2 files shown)


stp-ip (approved):
Great idea.
/lgtm

PR #869, opened 2022-02-16
Restore compatibility with Hugo v0.92.2

### LINKED ISSUE #868: Incorrect use of `ne` comparison breaks with hugo v0.92.2
bug

**What happened**:

After updating hugo to the newest version, executing `hugo server -D` caused the following error:

`ERROR 2022/02/14 20:12:17 render of "page" failed: execute of template failed: template: _default/single.html:5:6: executing "main" at <partial "helpers/fragments.html" (dict "page_scratch" $.Scratch)>: error calling partial: "[REDACTED]/themes/syna/layouts/partials/helpers/fragments.html:184:115": execute of template failed: template: partials/helpers/fragments.html:184:115: executing "partials/helpers/fragments.html" at <ne .Params.hide>: error calling ne: missing arguments for comparison`

No errors, as before the hugo update.

Check out the Syna starter repo and serve it with hugo v0.92.2:
```
git clone --recurse-submodules https://git.okkur.org/syna-start
cd syna-start 
hugo server -D 
```

This seems to be caused by gohugoio/hugo#9462. The change was released just a few days ago.

Changing `(ne .Params.hide)` to `(not .Params.hide)` in fragments.html:184 worked as a quick fix.

**Environment**:
- Syna Theme version:  v0.17.4
- Hugo version:  v0.92.2+extended darwin/amd64
- Others:

abrain:
This PR replaces the use of the `ne` template function with only one argument with an equivalent `not`. Hugo enforces this as of v0.92.2, which currently breaks the build of a Syna-based site on the latest versions.

fixes #868 

All the other uses of `ne` comparisons I could find, did correctly use 2 arguments.

**Release note**:

```release-note
Restore compatibility with Hugo v0.92.2
```

### DIFF
--- a/layouts/partials/helpers/fragments.html
+++ b/layouts/partials/helpers/fragments.html
@@ -181,7 +181,7 @@
       {{- $directory_same_name := in ($page_scratch.Get "fragments_directory_name") (printf "%s%s/" $root.File.Dir (replace $name ".md" "")) -}}
       {{- if and (not .Params.disabled) (isset .Params "fragment") (not $directory_same_name) -}}
         {{- if or $is_404 (ne .Params.fragment "404") -}}
-          {{- if and (eq .Params.fragment "config") (not (where ($page_scratch.Get "page_config") ".Name" $name)) (ne .Params.hide) -}}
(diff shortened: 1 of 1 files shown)


stp-ip (approved):
Thanks for the fix.

/lgtm

PR #871, opened 2022-03-11
Fix settings in copyright.html

yamanq:
This makes it possible to disable attribution with the `attribution` param and allows overriding the copyright message with `copyright`.

The settings are already tested for, but were not functional on hugo v0.93.2.

**Release note**:

```release-note
Fix overriding copyright message and disabling attribution footer
```


yamanq:
I misunderstood the proper way to set these params.

PR #874, opened 2022-05-07
Fix comparison for `summary` option in `list` fragment

Natureshadow:
For me, the comparison with boolean true did not work for the following frontmatter:

```toml
summary = true
```

The string comparison works.


stp-ip:
This is used throughout the code so shouldn't be an issue.
Will keep this open, but won't merge as this fix is more a temporary and local thing. We can't change this boolean to a string especially as we use booleans throughout the code.
Maybe Hugo changed something to make this break. :thinking:

PR #877, opened 2022-08-11
feat: add parallax on hero

### LINKED ISSUE #814: Add support for background images and simple parallax scrolling
feature

It would be nice to support the following features:

1. To use image for background of a fragment. This would be a generic feature used for any kind of fragment; content, items, portfolio, graph, etc. Something similar to what we can in the hero fragment. The solution would be, e.g., to use asset.image for background and to specify it in a special property, e.g., asset.use_as_background.
2. If there is a background image set for a fragment, there should be a way how to set a simple parallax scrolling effect and the ratio how fast/slow the background image is moving - from static to some percentage of scrolling the page. E.g. asset.parallax = 0 for background image not moving by vertical scrolling, asset.parallax = 50 for background image moving by 50% of scrolling,  or asset.parallax = 100 (default) for background image moving with the same speed as scrolled page. Also values more than 100 could be allowed.

sergiopena:
Add boolean to set the hero background image as fixed to create a parallax scroll effect.

Adds a nice parallax effect to the Hero partial

**Which issue this PR fixes** 
partially fixes #814

I started learning Hugo and this theme seven hours ago, so I literally don't know what I'm doing.

Just saw that it looked super easy to add a bool to fix the image in the background and get the effect 😇 

It works on my computer ™️ 

**Release note**:
```release-note
 parallax effect to hero partial
```

### DIFF
--- a/layouts/partials/fragments/hero.html
+++ b/layouts/partials/fragments/hero.html
@@ -12,6 +12,9 @@
         {{- with .Params.minHeight -}}
           {{- (printf "min-height: %s;" .) | safeCSS -}}
         {{- end -}}
+        {{- if .Params.parallax -}}
+          background-attachment: fixed;
+        {{- end -}}
         background-image:url({{ partial "helpers/image.html" (dict "root" $self "asset" .Params.header) }})"
       class="jumbotron text-center header-image mb-0
         {{- printf " bg-%s" $bg -}}


stp-ip:
Thank you for the addition already.
There are a few missing things.
It isn't documented on the fragment page: https://github.com/okkur/syna/tree/master/exampleSite/content/fragments/hero
Additionally with the default image it seems to not work well. Maybe a two part screenshot attached to the PR showcasing this effect makes this easier to grasp for anyone following up.

PR #883, opened 2022-09-23
Fix search for websites using subpaths in baseURL

### LINKED ISSUE #882: [Bug]: search does not work for websites with baseURL different than /
### What happened?

If a website using a `baseURL` like `https://example.com/foo` (i.e., using a subpath) enables the search fragment, a search triggers a `GET` request to `https://example.com/index.json`, which results in a `404` response and therefore in an empty search result regardless of the term queried by the user.

### What did you expect to happen?

I would expect a `GET` request to `https://example.com/foo/index.json`, which would allow the search fragment to work as expected.

### How can we reproduce this issue?

1. Set hugo `baseURL` in `config.toml` to something with a a subpath in the site url
2. Open website in browser of choice
3. Start network debugging feature in browser of choice
4. Trigger a search
5. Observe an empty search response on the site and a `404` response in the debugger

### Is there anything else we need to know about?

I'll try to provide a respective PR asap.

### Syna theme version

7fcac1f8

### Hugo version

v0.78.1

### Versions of other dependencies

_No response_

xai:
Fix search for websites using subpaths in baseURL.
Instead of always fetching `/index.json`, a potential subpath within the `baseURL` of a website should be considered.

fixes #882 

**Release note**:

```release-note
Fix search for websites using subpaths in baseURL
```


stp-ip (approved):
Thanks for including the fix.
/lgtm

PR #888, opened 2022-12-01
fix pagination

### LINKED ISSUE #887: [Bug]: Pagination des not work
### What happened?

Only the last 10 items in a list are displayed. There is not chance to see the older items.

### What did you expect to happen?

A pagination element should appear at the end of the list.

### How can we reproduce this issue?

Create a list with more than 10 items.

### Is there anything else we need to know about?

_No response_

### Syna theme version

_No response_

### Hugo version

_No response_

### Versions of other dependencies

_No response_

jze:
in the helper the variable is called `renderPagination` not `render_pagination`

fixes #887


stp-ip (approved):
Thanks for catching and fixing.
Retried deploy is working: https://deploy-preview-888--syna-demo.netlify.app/

/lgtm

